PR #5, opened 2021-01-27
separate meta data and data and join all data of a set of experiments…

joergfunger:
Separate the data generation process into
1. create metadata files
2. use metadata files to generate virtual raw data

The metadata files can then be used in the database to describe/query the actual raw data. I also combined all data files of specific virtual model test case (linear model with c=0 and second quadratic model with b=0) into a single hierarchical data file - even though they represent different experiments for different "a".


joergfunger:
Restructure into metadata and data.

joergfunger:
Separate data and metadata for the virtual experiments for triple store access.

PR #7, opened 2021-02-06
Add read the docs

joergfunger:
Started a general documentation using readthedocs.


TTitscher:
You could also have a look [here](https://github.com/BAMresearch/fenics-constitutive/blob/master/.github/workflows/python-package.yml#L24-L31) on how to add an automatic deploy of the website to (I guess) bamresearch.github.io/ModelCalibration. Is that planned? If so, I could also take care of that.

joergfunger:
I've already added the basic setup and a first introduction, the pull request is to merge this into master.

TTitscher (approved):
I just thought "github.io" would be more fancy ;)

But isn't there a part missing that tells readthedocs to actually run "doit website" and which htmls to pick?

Edit: nevermind, sorry, got it.

PR #8, opened 2021-02-07
Unger add correlation

joergfunger:
Restructure where latent variables are set and add bayesian inferences for the variables (correlation not yet included).


usecases/linearFunction/linear_model.py
```diff
@@ -34,6 +35,18 @@ def __call__(self, parameters):
...
+    def get_parameter_dict(self):
+        """Create a parameter list initialized with parameters all set either to None
+        or to some standard value (in case this is a deterministic variable, this value will
+        always be used
+
+        Returns: parameter list
+        """
+        prm = ModelParameters()
+        prm.define("a", None)
+        prm.define("b", None)
```
TTitscher, line 47: I first thought about just returning a list of ["a", "b"], but the actual ModelParameters with None is very nice. Would you agree if I, over at BayesianInference, change default value of
`def define(self, name, value=0)`
to
`def define(self, name, value=None)`
?

usecases/linearFunction/test_optimize.py
```diff
@@ -34,33 +37,48 @@ def __init__(self, yaml_file_metadata, yaml_file_data):
...
-            # add parameters that are model parameters (not given in the experimental data)
-            parameter.define("b")
+            # store the parameter list here to access it later, e.g. in the evaluation of the likelihood
+            self.linearModelParameters.append(parameter)
+            # add the parameter list with the model error to the joint latent list
             self.add(single_model_error, parameter)
 
-        # define one shared latent parameter (only one parameter b for all linear models)
-        self.latent.add_by_name('b')
+    def latent_as_tensor_variable(self, model, prior_pymc3):
```
TTitscher, line 53: I think, with your latest commit, you do not need this method anymore.

usecases/linearFunction/test_optimize.py
```diff
@@ -34,33 +37,48 @@ def __init__(self, yaml_file_metadata, yaml_file_data):
...
-
-        start_vector = np.array([0.7])
+            MultiLinearModelError('virtual_experiment_linear_meta.yaml',
+                                  'virtual_experiment_linear_data.yaml')
+        # define one shared latent parameter (only one parameter b for all linear models)
+        # add all variables 'b' in all model parameter lists as a shared latent variable
+        # return index in the global vector
+        index_b = all_experiments_model_error.latent.add_by_name('b')
+        start_vector = np.empty([len(all_experiments_model_error.latent)])
+        start_vector[index_b] = 0.7
```
TTitscher, line 81: Thank you for trying out my recommendation. But actually, I must admit that this also looks a bit clumsy. And it only makes sense if the "add_by_name" really returns a zero. I mean, if the _user_ must initialize the `start_vector` -- instead of just telling the library "Set the value of my latent parameter to 0.7" -- the "index_b" is rather pointless.

So ideally, we should IMO generalize the idea of this "UncorrelatedNormalPrior" (that takes care of setting the latent parameters _and_ the prior and, thus, hides the internal ordering from the user (if he wishes)) to any 'prior' (including pymc3, just numbers, MVNs,...)

usecases/linearFunction/test_optimize.py
```diff
@@ -120,18 +130,11 @@ def test_linear_with_noise_mcmc(self):
                                   'virtual_experiment_linear_with_noise_mcmc_data.yaml')
 
         # add noise term as local parameter to all model errors
-        #todo change latent in multimodelerror, since this is how we import the complete list
-        #todo do we have to add hyperparameters to the model parameters, since they are not part of the forward model
-        #howto access the parameter list outside (e.g. to add a model parameter)
-        for key, param_list in all_experiments_model_error.latent._all_parameters.items():
+        # currently we have to add hyperparameters to the model parameters, since they are not part of the forward model
+        for param_list in all_experiments_model_error.linearModelParameters:
```
TTitscher, line 134: I know, VB does not exist in this test case. But this problem could also be solved with VB, where the noise parameters must not be part of the parameter list. So in the long run, the this "joining" of parameters and hyperparameters for sampling methods must be done somewhere, but IMO later in the process. The initial definition should be independent of the inference method. 

That inspired https://github.com/BAMresearch/BayesianInference/issues/11

usecases/linearFunction/test_optimize.py
```diff
@@ -88,23 +106,127 @@ def weighted_all_experiments_model_error(prm_vector):
...
+
+        all_experiments_model_error = \
+            MultiLinearModelError('virtual_experiment_linear_with_noise_mcmc_meta.yaml',
+                                  'virtual_experiment_linear_with_noise_mcmc_data.yaml')
+
+        # add noise term as local parameter to all model errors
+        # currently we have to add hyperparameters to the model parameters, since they are not part of the forward model
+        for param_list in all_experiments_model_error.linearModelParameters:
+            param_list.define('sigma_noise_f', 1.)
+            param_list.define('sigma_noise_df', 1.)
```
TTitscher, line 136: I know, VB does not exist in this test case. But this problem could also be solved with VB, where the noise parameters must not be part of the parameter list. So in the long run, the this "joining" of parameters and hyperparameters for sampling methods must be done somewhere, but IMO later in the process. The initial definition should be independent of the inference method.

That inspired BAMresearch/BayesianInference#11

TTitscher (approved):
Despite a lot of meetings/discussions on the actual design, mostly outside of github, it is nice that put so much work into this and also another perspective. We will soon converge!

usecases/linearFunction/test_optimize.py
joergfunger: yes, thanks, I will delete that.

usecases/linearFunction/test_optimize.py
joergfunger: I would say that this strongly related to the question, if we should have an internal (related to the total latent) vector representation that we can access and extract as a vector (e.g. to be extracted and used [here](https://github.com/BAMresearch/ModelCalibration/blob/a6c52122b82b3499243d19fe654f8f67d44e3c70/usecases/linearFunction/test_optimize.py#L82.) Currently we do not have this (e.g. each entry in the latent vector might be related to different entries in the local vectors), and they are also just numbers - as opposed to e.g. a distribution (even a multivariate distribution might be required). As for the return of zero for add_by_name, I would say this is also works for more than one variable, since if your latent vector has 5 entries, each index would return the corresponding index that the variable is related to, and then you can directly access this entry in the latent variable via the index.

usecases/linearFunction/test_optimize.py
joergfunger: From my point of view, we would have to discuss what should be the interface. You can easily do VB without the need to add the hyperparameters to the model parameter list, though you might have to set up your VB more manual - you would have to define your complete inference problem twice, i.e. different procedures (just the model error is ientical) - e.g. the optimization test at the very beginning also does not use the hyperparameters either. The alternative would be to have a standard interface for the prior. But for the likelihood we have to define the two approaches twice anyway (pyMC3 uses directly the logLike, whereas in the currently implementation of VB the model error is directly used). You certainly would be able to come up with a standard loglike function for pyMC3 that corresponds to the VB implementation, but the inverse is not always possible, thus we would limit the applicability of pyMC3 to the problems solveable with our VB implementation. Maybe  it is be possible to allow both approaches (using a well defined loglike and prior with the same assumptions as in VB, or having everything completly independent).

usecases/linearFunction/test_optimize.py
joergfunger: see [here](https://github.com/BAMresearch/BayesianInference/issues/11#issuecomment-776735534)

usecases/linearFunction/linear_model.py
joergfunger: Yes, that prevents using variables that are not initialized.

PR #9, opened 2021-02-10
Steg

andreasstegmueller:
added two use cases by steg. Asymptotic model needs some debugging. Readmes tbd.


usecases/steg_asymptotic_2sensors/asymptotic_model_error.py
```diff
@@ -0,0 +1,82 @@
...
+        MatC = c + dx + ex²
+        and Mat_B_C = MatB / MatC
+        f(x): Mat_B_C = mA + mB/x
+
+    Attributes:
+        x_Mat_sensors (np.array): Array of positions x where Mat values are measured
+        data_MatB (np.array): Array of sensor readings at x_Mat_sensors
+        data_MatC (np.array): Array of sensor readings at x_Mat_sensors
+        data_Mat_B_C  : (np.array): Array of evaluated MatB / MatC data
+        asymptotic_model (AsymptoticModel): asymptotic forward model (mA+mB/x) , where 'mA' and 'mB' are model parameters
```
joergfunger, line 25: I'm trying to understand your problem. If you have measurements for matB and matC, you can define individual model error and inference procedures to get the individual parameters - resulting either in values for a,b,c,d,e or in distributions (when using Bayesian inference). If you have measurements for MatB, MatC as well as Mat_B_C, you can solve the inference/optimization problem for mA and mB with the model error
```
me_Mat_B_C = data_Mat_B_C - (mA+mB/x)
```
So those are all individual problems that seem to be unrelated at a first glance. Now, what might be the problem that your are solving (correct me if I'm wrong) - you only have measurements for MatB and MatC, and you would like to estimate what are the values mA and mB. In this case you can still generate the model errors for MatB and MatC independently and solve for a, b and c, d, e (or all together, if you don't care about the computational effort), and then solve another problem for Mat_B_C, where the data is now not generated from experiments, but is the (potentially noisy) output of models B and C. The joint problem with all variables a,b,c,d,e,mA,mB would look like this
```
me_MatB = data_MatB - (a+b*x)
me_MatC = data_MatC - (c+d*x+e*x**2
me_Mat_B_C = (a+b*x)/(c+d*x+e*x**2) - (mA +mB/x)
```

usecases/steg_asymptotic_2sensors/asymptotic_model_error.py
```diff
@@ -0,0 +1,82 @@
...
+        Args:
+            parameters(ModelParameters): a dictionary type of parameter list defined in bayes (a and b)
+
+        Returns:
+            concatenated model error (np.array) of all individual model errors
+        """
+        # [MatB, MatC, Mat_B_C] = self.asymptotic_model(parameters)
+        Mat_B_C = self.asymptotic_model(parameters)
+#        return np.concatenate((f-self.data_MatB, df-self.data_MatC))
+        return np.concatenate(Mat_B_C-(self.data_MatB / self.data_MatC) )
```
joergfunger, line 60: See above, what is the data that you have? If you have measurements for MatB  and MatC, they should also be included to fit you a, b, c.. The question here is really, whether you want to define your model error in Mat_B_C based on measurements data_MatB and data_MatC, or if you want to define your error based on the output of models B and C?

usecases/steg_asymptotic_2sensors/asymptotic_model_error.py
```diff
@@ -0,0 +1,82 @@
...
+        [Mat_B_C] = self.asymptotic_model(parameters)
+        return OrderedDict([('Mat_B_C', Mat_B_C-self.data_MatB/ self.data_MatC)])
+
+    def get_parameter_dict(self):
+        """Create a parameter list initialized with parameters given in the experimental data file
+
+        Returns: parameter list
+        """
+        prm = ModelParameters()
+        prm.define("a", self.a) # old
```
joergfunger, line 80: see above, are you fitting mA and mB, or a, b, c, d, e, or all of them together.

usecases/steg_asymptotic_2sensors/virtual_experiment.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+        with open(virtual_experimental_data_file, "w") as f:
+            data = {}
+            MatB_x = self.a * np.ones(len(self.x_Mat_sensors)) \
+                    + self.b * self.x_Mat_sensors \
+                    + np.random.normal(0, self.sigma_noise_Mat, len(self.x_Mat_sensors))
+            MatC_x = self.c * np.ones(len(self.x_Mat_sensors)) \
+                    + self.d*  self.x_Mat_sensors  + self.c * np.square(self.x_Mat_sensors) \
+                    + np.random.normal(0, self.sigma_noise_Mat, len(self.x_Mat_sensors))
+            Mat_B_C_x = MatB_x / MatC_x
```
joergfunger, line 55: So do you have data for all intermediate steps. In this case, I would argue you have three independent optimization problems, because if you know what the output of B and C is then you can use this information directly in the model for B_C instead of relying on outputs of the models B and C.

joergfunger (changes requested):
Thanks for the first merge request. I've added some comments, in particular the problem to be solved, and the data that is used, has to be clarified.

usecases/steg_asymptotic_2sensors/asymptotic_model_error.py
andreasstegmueller: Jörg, thanks for your comments - and apologize for not replying earlier, I simply overlooked them..

You are correctly seeing my problem. Both variants are relevant to our models in general, but in this use case I have measured (virtual) data for MatB and MatC. Mat_B_C is not measured but a direct evaluation of the prior data series. (These two series might be represented by models with the parameters a,b and c,d,e.)
For now, I am not interested in models of MatB and MatC, resepctively, but only in the model for Mat_B_C with the parameters mA and mB (the "target property"). In principle, this asymptotic model can be generated from experimental data (data_MatB / data_MatC) or from separate models with a,b and c,d,e (thus, two options). As said, for now, only the last model error definition me_Mat_B_C is relevant.

usecases/steg_asymptotic_2sensors/asymptotic_model_error.py
andreasstegmueller: Here was my problem, I think. Thank you.

usecases/steg_asymptotic_2sensors/virtual_experiment.py
andreasstegmueller: This was also wrong, I think. I have to you the "measured" data_MatB and data_MatC...

usecases/steg_asymptotic_2sensors/asymptotic_model_error.py
andreasstegmueller: > 
> 
> Jörg, thanks for your comments - and apologize for not replying earlier, I simply overlooked them..
> 
> You are correctly seeing my problem. Both variants are relevant to our models in general, but in this use case I have measured (virtual) data for MatB and MatC. Mat_B_C is not measured but a direct evaluation of the prior data series. (These two series might be represented by models with the parameters a,b and c,d,e.)
> For now, I am not interested in models of MatB and MatC, resepctively, but only in the model for Mat_B_C with the parameters mA and mB (the "target property"). In principle, this asymptotic model can be generated from experimental data (data_MatB / data_MatC) or from separate models with a,b and c,d,e (thus, two options). As said, for now, only the model error definition me_Mat_B_C = data_MatB / data_MatC is relevant. (What I have not addressed: data_MatB and data_MatC, if measured, might have different data uncertainties, respectively...)

usecases/steg_asymptotic_2sensors/asymptotic_model_error.py
andreasstegmueller: I am fitting mA and mB.

usecases/steg_asymptotic_2sensors/virtual_experiment_asymptotic_model_data.yaml
```diff
@@ -0,0 +1,15 @@
+MatB: [17.0, 40.4473684210526, 63.89473684210526, 87.34210526315792, 110.78947368421052,
```
joergfunger: Please remove th yaml files from the git repo, they can always be regenerated on the fly. In particular, if you change someting in the input, you would have to commit the changes to the yaml files, which is suboptimal and error prone.

usecases/steg_asymptotic_2sensors/virtual_experiment_asymptotic_model_meta.yaml
```diff
@@ -0,0 +1,11 @@
+a: -280
```
joergfunger: same here, please remove. Please also remove the binary files (*.pyc)

usecases/steg_asymptotic_2sensors/asymptotic_model_error.py
joergfunger: see above then remove data_Mat_B_C, since it is not used at all.

usecases/steg_asymptotic_2sensors/asymptotic_model_error.py
joergfunger: This is still wrong here, if you are fitting mA and mB, a is never used and should be removed here. And I suggest defining the model parameters in the forward model (in the current version that's correctly done, but I changed that in the update here - both versions work, but I think it is more straightforward the the model returns it's parameters rather than the model error)
[updated version in another merge request](https://github.com/BAMresearch/ModelCalibration/blob/unger_add_correlation/usecases/linearFunction/linear_model.py)

usecases/steg_asymptotic_2sensors/generate_metadata_virtual_experiment.py
```diff
@@ -0,0 +1,69 @@
+import numpy as np
+import yaml
+
+
+def write_metadata_to_yaml(virtual_experimental_metadata_file,
+                           a, b, c, d, e,
```
joergfunger: if you want to prescribe in the experiment a, b, .. instead of Mat_B and mat_C then you could add the definition of the models here (just move that from the forward model).

joergfunger: I've also changed in this correlation request to store the files in json, it's possible to keep this here as is, just want to mention this.

usecases/steg_asymptotic_2sensors/test_optimize.py
```diff
@@ -0,0 +1,129 @@
...
+            # read in the parameters that are given in the experiment (in this case constant offset a)
+        parameter = single_model_error.get_parameter_dict()
+            # add parameters that are model parameters (not given in the experimental data)
+        parameter.define("mA")
+        parameter.define("mB") #! was b
+        self.add(single_model_error, parameter)
+
+        # define one shared latent parameter (only one parameter for all models)
+        self.latent.add_by_name('mA')
+#        self.latent.add_by_name('mB')
```
joergfunger: so do you assume mB to be known? In this case, you would have to define a value to mB at some point.

usecases/steg_asymptotic_2sensors/test_optimize.py
```diff
@@ -0,0 +1,129 @@
...
+            all_model_errors = all_experiments_model_error.evaluate(prm_vector)
+            obj = 0.
+            for key, me in all_model_errors.items():
+                # obj = obj + np.dot(me['MatB'], me['MatB']) / d['sigma_noise_Mat'] ** 2 + \
+               #       np.dot(me['MatC'], me['MatC']) / d['sigma_noise_Mat'] ** 2 \
+                #         +  np.dot(me['Mat_B_C'], me['Mat_B_C']) / d['sigma_noise_Mat'] ** 2
+                obj = obj + np.dot(me['Mat_B_C'], me['Mat_B_C']) / d['sigma_noise_Mat'] ** 2 
+            return np.sqrt(obj)
+
+        weighted_start_vector = np.array([0.7])
```
joergfunger: did you check that the start vector is somehow close to your exact values? Since this is a local iteration that should match. For extending this to a Bayesian procedure, we need priors that similar reflect prior knowledge about our parameters.

usecases/steg_asymptotic_2sensors/asymptotic_model_error.py
joergfunger: So in your current implementation, I'm not sure why you need the parameters a, b, c, d, e at all. I would suggest to either remove those, or extend the model error such that you have your second option. In the first option, you essentially have data for Mat_B_C mat_B and mat_C and try to evaluate mA and mB in your combined model - this is essentially a fitting of a function Mat_BC = mA+mB/x with data pairs  Mat_B_C = mA/mB. So I don't understand your measurents, because you have now measured that twice (once Mat_B_C directly and once mA and mB that allow you to compute Mat_B_C). So in that scenario, I think Mat_B_C should not be a measurement and removed from the data.

joergfunger: I personally think though that your second option is much more interesting. In this case, you have measurements for all three, and Mat_B_C is directly a function of the model output of matB and matC, thus the model error is in Mat_B_C is not computed using the measured data, but using the models matB and matC.

joergfunger (changes requested):
Thanks for the new update. I think we are converging. There are some minor issues added to the comments. Please also remove your second case. In addition, we should have a short discussion again on what is you data and what are the models, since I think I start understanding what you want to do, but I think this is not yet what is in the code.

PR #12, opened 2021-07-12
Add files via upload

MathisNo:
Added the Ontologies as well as the script to create the graph for the concrete-usecase


aklawonn (approved):
Generally, I would suggest using no German file or directory names, and avoid using white spaces in those names. Also, I would suggest using a common naming convention. For example, instead of having paths like "usecases/Concrete/Data", I would go with "usecases/concrete/data". However, I don't want to break any scripts that rely on these names, but in the future we should converge to something that looks less messy with respect to the used file/directory names.

PR #14, opened 2022-02-24
Shacl validation action

horstf:


usecases/Concrete/knowledgeGraph/shape.ttl
```diff
@@ -0,0 +1,56 @@
...
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+@prefix mseo: <https://purl.matolab.org/mseo/mid/> .
+@prefix cco: <http://www.ontologyrepository.com/CommonCoreOntologies/> .
+@prefix obo: <http://purl.obolibrary.org/obo/> .
+@prefix con: <http://w3id.org/concrete> .
+@prefix tt: <http://w3id.org/concrete/youngs> .
+
+schema:SpecimenDiameter
```
joergfunger, line 22: Could you add some documentation here, such that someone not familiar with the definitions is able to understand what is going on. In addition, a link to a general documentation on the structure of the shacl shapes would be great.

joergfunger (commented):
Thanks, couple of remarks, in case of questions, let me know. Otherwise, it would be great to revise the merge request and then send again.

horstf:
I have refactored the steps into a python script which is now called in the workflow and also incorporated it into the dodo.py.

horstf:
As for tests of the shapes, @firmao seems to be working on concrete shacl shapes for the data so I think it would make sense to test them once, after they are developed.

firmao:
> As for tests of the shapes, @firmao seems to be working on concrete shacl shapes for the data so I think it would make sense to test them once, after they are developed.

First of all, thanks for the comments!

Indeed, make sense to test them once, after they are developed, let me know if I could help.

usecases/Concrete/knowledgeGraph/emodul/emodul_validation.py
```diff
@@ -0,0 +1,57 @@
...
+baseDir2 = Path(__file__).resolve().parents[2]
+ontologyPath = os.path.join(baseDir2,'ConcreteOntology')
+metadataPath = os.path.join(baseDir0,'E-modul-processed-data/emodul_metadata.csv')
+graphPath = os.path.join(baseDir0,'E-modul-processed-data/EM_Graph.ttl')
+#importedOntologiesPath = os.path.join(baseDir2,'GraphCreation/Ontologies')
+processedDataPath = os.path.join(baseDir0,'E-modul-processed-data')
+
+# clone and run rdfconverter docker
+Repo.clone_from('https://github.com/Mat-O-Lab/RDFConverter.git', os.path.join(baseDir1, 'RDFConverter'))
+subprocess.run(['docker-compose', '-f', os.path.join(baseDir1, 'RDFConverter/docker-compose.yml'), 'up', '-d', 'app'], stderr=subprocess.DEVNULL, stdout=subprocess.DEVNULL)
```
joergfunger, line 21: where is the docker-compose requirement set? IMO, that would be required in the conda environment?

usecases/Concrete/knowledgeGraph/emodul/emodul_validation.py
```diff
@@ -0,0 +1,57 @@
...
+
+passed = [] #list of filenames of files that passed validation
+failed = [] #list of tuples (filenames, reports) of files that failed validaton
+
+# read shacl shape
+with open(os.path.join(baseDir1, 'shape.ttl'), 'r') as f:
+    shacl_shape = f.read()
+
+sleep(3)
+s = requests.session()
```
joergfunger, line 31: IMO it would be great to have some comments on this interface (what the the docker container starts a server that is accessed via https ...)

usecases/Concrete/knowledgeGraph/emodul/emodul_validation.py
```diff
@@ -0,0 +1,57 @@
...
+        passed.append(rdf_file)
+    else:
+        failed.append((rdf_file, r.text))
+
+# close session object and shutdown docker container
+s.close()
+subprocess.run(['docker-compose', '-f', os.path.join(baseDir1, 'RDFConverter/docker-compose.yml'), 'down'], stderr=subprocess.DEVNULL, stdout=subprocess.DEVNULL)
+
+# print passed files
+print('files that passed shacl validation:')
```
joergfunger, line 49: There should not be a print, but a check for files that have successfully passed the pytest.

usecases/Concrete/knowledgeGraph/emodul/emodul_validation.py
```diff
@@ -0,0 +1,57 @@
...
+for t in enumerate(passed, start=1):
+    print(f'{t[0]}. {t[1]}')
+
+# print failed files
+print('files that failed shacl validation:')
+for t in enumerate(failed, start=1):
+    print(f'{t[0]}. {t[1][0]}\n{t[1][1]}')
```
joergfunger, line 56: Same here, it would be good to have a failed file and directly check for the identified problem (e.g. no author set), and return true in pytest, if this error message is returned.

joergfunger (commented):
All the changes are accepted. There is only one detail, and maybe that is something we haven't looked into in general yet, but the tests should not print a value, but directly have asserts (that are even for failed tests true), e.g. assert that a specific exception has be thrown, or assert that the shacl validator return a specific triple indicating a missing/failed shacl constraint.

usecases/Concrete/knowledgeGraph/emodul/emodul_validation.py
firmao: Hi @joergfunger, about "check for files that have successfully passed the pytest"
Could you please give an example of the output or something else that satisfies you?

joergfunger: So of I write a standard test that returns a result (e.g. that there is not violation of the constraints), then we should check for this in the test (so not just printing the result, but adding assertions https://docs.pytest.org/en/6.2.x/example/simple.html). Same when something might go wrong, so the constraints are not fulfilled. The function then returns the failed constraints, and we can add assertions such that the test checks that exactly these failed constraints (for a given data set that does not fulfill the constraints) are returned.

joergfunger: Since this is a test environment, the dataset does not change. This is not supposed to be used dynamically for any data set, but this workflow should rather be a workflow test that demonstrates the features.

horstf: https://github.com/BAMresearch/ModelCalibration/blob/shacl-validation-action/usecases/Concrete/knowledgeGraph/emodul/emodul_validation.py#L101-L107

Thank you, this explanation helped. So I've changed a lot of the validation code now and we can now test that a specific shape of the shapes graph has failed or not. As shown in the lines of code linked above. We can then define shapes in the shapes graph that should fail or that should not fail and check that they do (or don't) via pytest.

PR #18, opened 2022-04-13
Create python module to validate processed data

horstf:


joergfunger (commented):
Not sure if the merge request is uptodate, at least test files are currently not working.

horstf:
> @horstf we have updated the structure, could you try to adapt your code the way it is now working in this new structure, such that we could merge it into main?

Is that the structure in the /lebedigital folder? I.e. refactor my code in an appropriate /lebedigital/validation folder?

horstf:
I've copied the validation submodule into the new lebedigital package. I haven't updated the dodo.py file yet because the current version is not using the new package for now.

lebedigital/validation.py
```diff
@@ -0,0 +1,79 @@
+from pyshacl import validate
+from rdflib import Graph, URIRef, Namespace
+from rdflib.util import guess_format
+from rdflib.namespace import SH, RDF
+
+"""
+baseDir0 = Path(__file__).resolve().parents[0]
```
joergfunger, line 7: Is that comment still needed, or could that be removed?

lebedigital/validation.py
```diff
@@ -0,0 +1,79 @@
...
+
+SCHEMA = Namespace('http://schema.org/')
+
+"""
+Given a path to a shacl shape and a path to an rdf file, this function tests the rdf data against the specified shacl shapes.
+The result is an rdflib graph containing the validation report, if it is empty the validation was successful.
+"""
+def test_graph(rdf_graph: Graph, shapes_graph: Graph) -> Graph:
+
+    conforms, result_graph, _ = validate(
```
joergfunger, line 24: could you document the input parameters, see e.g. https://github.com/BAMresearch/probeye/blob/main/probeye/postprocessing/sampling_plots.py

usecases/Concrete/knowledgeGraph/emodul/validation.py
```diff
@@ -0,0 +1,78 @@
+from pyshacl import validate
```
joergfunger, line 1: not sure that that file does, but all files should eiter be in lebedigital (the actual functions), in tests (the tests for these functions) or in minimumWorkingExample (the pydoit workflow that processes the data for the minimum working example). The folder usecases/Concrete is deprecated.

joergfunger (changes requested):
thanks for the pull request, there are some modifications required to adapt to the new structure and some documentation issues.

lebedigital/validation.py
horstf: I've added the comments

usecases/Concrete/knowledgeGraph/emodul/validation.py
horstf: That file is not used anymore, I think that was an accidental commit

joergfunger: This file is still in Concrete, please remove it?

usecases/Concrete/knowledgeGraph/requirements.txt
```diff
@@ -8,5 +8,6 @@ SPARQLWrapper==1.8.5
 requests==2.22.0
```
joergfunger, line 8: This file is also outdated, the conda environment at the top level is the one and only to be used.

joergfunger (changes requested):
Thanks for the changes, I still have a few comments. Where are the tests related to validation.py? In addition, the folder usecases/Concrete is deprecated.

usecases/Concrete/knowledgeGraph/requirements.txt
horstf: We can ignore these commits then, all the libraries I need are already part of the environment file.

horstf: I've also deleted the validation.py from Concrete

horstf: For the tests, is there a graph that I can use to test my code? The usecases/MinimumWorkingExample/emodul/processed_data folder is empty but I would need an EM_Graph.ttl file or so to test the validation. I can also just create my own otherwise from the data.

joergfunger: Could you talk to @PoNeYvIf  on that? For the tests, it would also be possible (and maybe even better) to create your own very simple test cases (it does not have to be related to a very specific ontology, but rather a very general one such as foaf or prov). In particular make sure that there are errors existing such that those can be returned and processed. We could also have a short phone call on that.

horstf: @joergfunger Okay, I've added a simple graph and some shapes to test it against. I've added the files to usecases/MinimumWorkingExample/emodul/validation_test/ but I'm not sure if that is the right directory for the test data files. I can move them if another location would be better. I've also added the file tests/validation/test_validation.py which tests the shapes against the graph and checks that some shapes fail and others don't.

joergfunger: For the validation, this would work. The only thing remaining before we can merge is the adaption of the doit script in usecases/minimumWorkingExample (thus not dodo.py in Concrete) and the removal of the changes related to the requirements in Concrete.

horstf: I've removed the changes in requirements, im not sure how the adoption for the doit script should look like, should i talk to @PoNeYvIf or should we schedule a meeting maybe?

joergfunger: You can have a look [here](https://github.com/BAMresearch/LebeDigital/blob/main/usecases/MinimumWorkingExample/dodo.py). The final task loops over all the processed data files and generates corresponding ttl files. We would then do another validation task that loops over all the ttl files and validates them. If there are further questions, let's have a short phone call.

joergfunger: The only thing that is currently not completly correct in this current dodo file (@PoNeYvIf) is the dependency on the metadatafiles. They are not existing when building the workflow graph and does might lead to problems for the first execution. Thus, there are two options, one is a decorator after (I think @eriktamsen) wanted to change that, and the second option is making the ttl file dependent on the original raw data file that is always present (thus not running into the problem that they are not existing at the very first run).

eriktamsen: As Jörg said, those two options exist. I successfully used the decorator option in a different branch: https://github.com/BAMresearch/LebeDigital/blob/44-getting-data-from-kg-for-calibration/usecases/MinimumWorkingExample/dodo.py
Note, I also added there the option of a "cheap" workflow as the default, which only processes a single file (we might change it in the future once we can asseble the whole KG), mostly because the new calibration step is too expensive to run every time for every file. If you have questions, let me know.

horstf: @joergfunger Sorry for the late reply. Okay, so I would loop over all ttl files and check the validation for each, what is the output then? Should I abort the script if one validation fails? Or should I output the results in a text file?

joergfunger: The would be that we iteratively generate a knowledgegraph, i.e. we have one large triple store with all the existing triples. Now we would add (in our case this is the loop over all individual ttl files) the data for a new test. @PoNeYvIf is currently trying to do this. Before we actually add each individual file to the triple store, we would check the correctness of the data set, i.e. using the shacl validation. That said, we should have one loop over all raw data files. In this loop, there have to be the substeps 
1. create meta_data.yaml (extract the meta data from the raw data)
2. create knowledge graph of metadata.yaml using the ontology
3. shacl-validate the (not uploaded) knowledge graph for the single experiment
4. if successful, add to global knowledge graph (in reality this would be the triple store, right now it is just our own joint ttl file)
So the subsequent addition to the knowledge graph is only done if it the check is sucessful, otherwise exit the procudure with an error message and a file the logs the shacl error graph.

One thing that I'm not sure how to do that in shacl is the question of checking for instances that are refered to in one graph but defined in another. (e.g. you have a machine whose description is already added to the global triple store, now you would like to add a compression tests which refers to this machine by an id, but we somehow want to test that this id - i.e. an instance of testing machine - is actually existing in the global KG). Is that possible with shacl?

horstf: So it would make sense to make the task fail if the validation does not succeed so that the next step (which would be the adding to the global KG) is not executed. 

As for the information stored in the global KG, we could read graphs from remote triple stores and join them with local graphs, this would mean some overhead though with copying the global KG every time we want to add to it. If we always check for the same properties with shacl we could write a SPARQL query that gathers the relevant data from the global KG based on the data in the local KG. Maybe @firmao can help?

horstf: E.g. if we know that we test for machine ids via shacl, we could get the machine IDs from the global KG that are referred to in the local KG. This would of course mean that every time we add a rule we have to think about the SPARQL query and the information that could be contained globally and not locally, i don't know if this is feasible.

firmao: Yes, let's try to create the queries, let me know when it's okay for you.

joergfunger: Not sure if I would always download the complete global graph. It would rather be necessary if particular instances that are referred to in the local graph (e.g. the testing machine that has already be created, or the mix_design we are referring to when adding a new Youngs modulus test) is existing. So for the specific tests, we would first generate the mix design, and upload it to the global KG. Then we would create a Youngs modulus test that references an instance of the mix design (without creating the same instance again). By uploading that then to the KG, these two data sets (mix design and Youngs modulus test) are automatically connected. And before uploading the Youngs modulus KG, we would have to check, if this instance of the mix design is already existing in the global KG (and potentially further properties apart from the id, but that is not necessary right now).

joergfunger: As for the sparql query, we somehow now in the local KG generation what classes we expect. So we the sparql query should return the ids of all instances of a given class that should be quite general and would not require rewriting the sparql query each time we have a new test.

horstf: Maybe we should have a talk together with @firmao to discuss the specifics?

firmao:
Sure, please, send a message on teams.

joergfunger:
@horstf What is the status with this merge request? It seems that only the integration into the dodo file in minimum working example is missing.

joergfunger:
@horstf  The actions are not working, so that prevents me from merging the branch. In addition, I checked the other files. I guess there was a misunderstanding with the dodo.py, since you changed that in concrete. I guess we would have to remove this directory (it was already deprecated, but not removed, because it included all the old stuff). The actual dodo.py we are working with is in usecases/MinimumWorkingExample. Could you update that there as well. I guess the problem is that we do not have a working generation of the KG there yet, @AidaZt @soudehMasoudian @alFrie how are we going to handle this merge request? Or should @horstf integrate that in pydoit without actually including in the tasks, such that you could add that afterwards once the KG is up and running?

horstf:
Hi, is there any update on how i should proceed?

joergfunger:
> 

Could we have a short teams meeting? Maybe that is easier

horstf:
> > 
> 
> Could we have a short teams meeting? Maybe that is easier

Yes, the date you suggested in your teams invitation works fine.

PR #21, opened 2022-04-29
Unittest antology development

soudehMasoudian:
I've written unittest related to emodul_metadata_extraction. Please check and merge it if everything is as it should be.
The only changes are a test folder in 'usecase/Concrete' path, and a script which named test_generate_metadata_from_emodul_BAM.py

It can be run by below command:
python -m unittest discover


usecases/Concrete/dodo.py
```diff
@@ -45,11 +45,21 @@ def task_emodul():
     }
     yield {
         'basename': 'extract emodul metadata',
-        'actions': ['python knowledgeGraph/emodul/emodul_metadata_extraction.py']
+        'actions': ['python knowledgeGraph/emodul/emodul_metadata_extraction.py'],
+        'targets': ['knowledgeGraph/emodul/E-modul-processed-data/emodul_metadata.csv']
+    }
+    yield {
+        'basename': 'calculate emodul',
+        'actions': ['python knowledgeGraph/emodul/emodul_calculation.py'],
```
joergfunger, line 53: not sure why we calculate the E-modulus here already, that is IMO part of the processing of the data afterwards.

usecases/Concrete/knowledgeGraph/emodul/emodul_calculation.py
```diff
@@ -0,0 +1,63 @@
...
+    minIndex = find_third_element_smaller_than_number(minForces, meanMinForces) 
+
+    specimenDiameter = df_metadata.loc[df_metadata['experiment raw name'] == file]['diameter'].values[0]
+    specimenDiameter = float(specimenDiameter.replace(',','.'))
+
+    specimenLength = df_metadata.loc[df_metadata['experiment raw name'] == file]['length'].values[0]
+    specimenLength = float(specimenLength.replace(',','.'))
+    
+    specimenSurfaceArea = math.pi * specimenDiameter**2 / 4000000
+    deltaSigma = (maxForces[maxIndex] - minForces[minIndex])*1000/specimenSurfaceArea
```
joergfunger, line 55: This computation is IMO wrong, the third cycle should be extracted. This would also be part of a data processing pipeline ind should be documented (in the knowledgegraph, there has to be an entry that specifies, which script (on the file server) was used to compute the Youngs modulus

usecases/Concrete/tests/test_generate_metadata_from_emodul_BAM.py
```diff
@@ -0,0 +1,124 @@
...
+import os
+from pathlib import Path
+from datetime import datetime
+
+class test_generate_metadata_from_emodul_BAM(unittest.TestCase):
+    
+    # read specimen.dat file to make sure that headers work correctly
+    def read_specimen_file(self, file):
+        with open(file) as specimen:
+            print ('test')
```
joergfunger, line 12: remove test stuff

usecases/Concrete/tests/test_generate_metadata_from_emodul_BAM.py
```diff
@@ -0,0 +1,124 @@
...
+            for i, line in enumerate(specimen):
+                if (specimen.name.__contains__('BA Los M VI-6')):
+                    if ("Bediener" in line):
+                        date_val = line.split('\t')
+                        origin_date = date_val[9]
+                        str_date = ''.join(origin_date.split())
+                        str_date_time = datetime.strptime(str_date, '%d.%m.%Y%H:%M:%S')
+                        self.assertIsInstance(str_date_time, datetime)
+                        self.assertListEqual([str_date_time.year, str_date_time.month, str_date_time.day, 
+                        str_date_time.hour, str_date_time.minute, str_date_time.second],
```
joergfunger, line 24: What are you testing here? The result that you compare with should not be obtained by reading in the file again, but take the result that is processed in the metadata extraction functions (those are the ones we would like to test).

joergfunger (changes requested):
There are a couple of question. 

- file_dep in pydoit should be added. You would have to specify not only the target, but also what it is required to compute this (and thus when this changes, the target needs to be updated)
- the test files should test the functions to extract metadata/compute parameters etc.. This is currently not done, but it is extracted again.

eriktamsen:
This looks to me as outdated and not based on the new extraction scripts, which have working tests.
In addtion we are now using pytest.
Can the branch be deleted?

PR #22, opened 2022-05-10
added three point bending test example

eriktamsen:
I created an example of a predictive simulation for the three point bending test.


atulag0711 (commented):
In general everything looks fine to me. Can be merged. As soon as the "knowledge graph not working issue" is fixed, i will test it also.

atulag0711 (approved):

PR #25, opened 2022-05-31
24 tests are not working

joergfunger:
I added a unique conda environment. There are still a couple of changes required, in particular the src should be moved outside of the examples and the copied files shoulld be removed.


PoNeYvIf (approved):
Good work! The changes make the file more readable. The new lebedigital.yml environment is really easy to deploy even if you have no knowledge of conda.

PR #33, opened 2022-07-05
Change doit in minimum working example

eriktamsen:
Metadata extraction is working, included in doit and a respective test is added.
The CI still has problems with the environment but we will fix that later.


eriktamsen:
Just as a general information, apparently there is a problem with conda-forg, which probably causes most of the errors that we have seen in the actions on Monday. I currently have the same problem e.g. if I want to install fenics in an empty environment locally. I dont think there is much else to do except wait till something gets fixed...

atulag0711 (approved):
Merging it for the time being so that we can process with this. The test are failing because of Fenics Concrete which Erik is fixing

PR #39, opened 2022-07-21
adding extraction script for mixture metadata

alFrie:
I added the metadata-extraction script for the mixture set to the new structure.


joergfunger (changes requested):
The file looks good to me, no comments. Though I would actually like to add only new functions that are actually tested. I guess you have already done that when developing the code. Maybe you could send that to @soudehMasoudian so that she can write the test in that branch here as well and we are only merging this into main once it is actually tested.

joergfunger:
Yes, just change the excel sheets. This is not the focus of the project. And then send your tests to @soudehMasoudian such that we could actually integrate the tests directly before merging into main.

lebedigital/raw_data_processing/metadata_extraction/mixture_metadata_extraction.py
```diff
@@ -1,142 +1,190 @@
...
     """
     
+
+    # Find sheets in the file containing the mixture (keyword: "Rezeptur")
     excelsheet = os.path.basename(locationOfRawData)
     print("\n Working on: "+ excelsheet)
     excelfile = pd.read_excel(locationOfRawData, sheet_name= None) 
     listofkeys = [i for i in excelfile.keys() if "Rezeptur" in i] 
-    print("Relevant sheets in this file: " + str(listofkeys))
+    print("Sheets containing mixture metadata in this file: " + str(listofkeys))
```
joergfunger, line 72: I think we should add a logger (e.g. loguru as used [here](https://github.com/BAMresearch/probeye/blob/6272085e8c69131b43b6a90c81f502636d3aa59d/probeye/definition/forward_model.py#L5) not to clutter the output with debugging information. An simple alternative would be to define a debug=True variable at the top and put that print statements into an if statement.

lebedigital/raw_data_processing/metadata_extraction/mixture_metadata_extraction.py
```diff
@@ -1,142 +1,190 @@
...
+        # operator name (always the same position)
+        metadata['tester_name'] = exceltodf.iloc[0,10]
+
+        # name of specimen
+        idx = labelidx["Bezeichnung der Proben:"]
+        metadata['specimen_name'] = exceltodf.iloc[idx,3] #4
+
+        # water cement ratio ("Wasserzementwert")
+        idx = labelidx["Wasserzementwert"]
+        metadata["watercementratio"] = exceltodf.iloc[idx,2] #7
```
joergfunger, line 116: what are these numbers #7 as comments?

joergfunger (approved):
Thanks, almost ready to merge. IMO, there is only one thing to change related to printing the progress. Then (and that is only a remark), we will have to adapt the procedure to our mix in the demonstrator, because the cement content is there actually decomposed into two different components (cemI and cemII) that we would have to store separately, but lets see how this is represented in the ontology and then we could change that afterwards. This is then another issue to be solved later.

eriktamsen (changes requested):
Thank you for the changes in the script. This look good.
Please move the script to a different folder as suggested in #55 and already implemented in main.

lebedigital/raw_data_processing/metadata_extraction/mixture_metadata_extraction.py
joergfunger: I would recompute it.

joergfunger: Just use the ratio between mass of water and mass of cement both given in the excel sheet separately. That should evaluate to a numeric value somewhere between 0.2 and 0.7 and should agree with the entry in this cell (without the additional sometimes weird comments that we currently do not use anyway).

lebedigital/raw_data_processing/metadata_extraction/mixture_metadata_extraction.py
```diff
@@ -0,0 +1,190 @@
...
+
+from cmath import nan
+import pandas as pd
+import os
+import yaml
+
+# function to convert german formatting to english
+def replace_comma(string):
+    if "---" in string:
+        string = nan
```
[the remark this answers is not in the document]
eriktamsen: I dont have an elegant solution. I would define a function e.g.:
def my_float(string):
   if string:
      string = float(string)
   
   return string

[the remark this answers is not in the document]
eriktamsen: I guess you are right. I did not know that "nan" is a possible float value. Should be good enough.
Ignore my comment ;).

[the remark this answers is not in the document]
joergfunger: Instead of adding a NaN or None, could we just don't include that key-value pair in the file? Then there is only valid data in the yaml-file.

joergfunger:
Can you join the two pull request (unit tests + metadata extraction).

joergfunger:
> > Please move the script to a different folder as suggested in #55 and already implemented in main.
> 
> I can't do that while still working on my branch, right? This new folder didn't exist in the main branch when I created mine. Is there a way to change this, should I create a new branch or do I have to move it when it is merged with main? @eriktamsen

You can merge the main branch into your branch, perform all the changes and then merge back. After having merged main into your feature branch, the directory structure should be updated.

lebedigital/raw_data_processing/youngs_modulus_data/mixture_metadata_extraction.py
```diff
@@ -0,0 +1,283 @@
...
+    elif format == 'float':
+        string = string.replace(',', '.')
+        return float(string)
+    else:
+        string = string.replace(',', '.')
+        return string
+
+
+# function to check for nan-values independently of the format (str/float)
+def isNaN(num):
```
joergfunger, line 48: why not use standard functions such as [this](https://numpy.org/doc/stable/reference/generated/numpy.isnan.html)

lebedigital/raw_data_processing/youngs_modulus_data/mixture_metadata_extraction.py
```diff
@@ -0,0 +1,283 @@
+# Script for metadata-extraction:
```
joergfunger, line 1: We should move the file to mixture (instead of Youngs modulus), because the mixture is not related to a specific (Youngs) modulus test.

joergfunger (approved):
Just very minor changes required, in particular moving the file since the mixture is general and not only for Youngs-modulus tests. Thanks a lot, looks very nice.

lebedigital/raw_data_processing/youngs_modulus_data/mixture_metadata_extraction.py
joergfunger: no, that has to be created

eriktamsen (approved):
Look great, thanks.

PR #43, opened 2022-08-26
Mapping script using only owl ready

PoNeYvIf:
The mapping script seems to be working as we want. However I don't quite understand the data so if you notice that something is wrong it's probably because I misunderstood something. In the test file there are also examples on how to query the knowledge graph that is generated.

### DIFF
--- a/lebedigital/mapping/emodul_mapping.py
+++ b/lebedigital/mapping/emodul_mapping.py
@@ -9,10 +9,13 @@ import arviz as az
 from owlready2 import *
 from yaml.loader import SafeLoader
 from rdflib import URIRef, Graph, Literal, BNode
-from rdflib.namespace import FOAF, RDF
+from rdflib.namespace import FOAF, RDF, XSD
+from pathlib import Path
 
-myDir = os.path.dirname(__file__)
-owlPath = os.path.join( myDir, "lebedigital", "ConcreteOntology")
+# myDir = os.path.dirname(__file__)
+# owlPath = os.path.join( myDir, "lebedigital", "ConcreteOntology")
+
+baseDir = Path(__file__).parents[2]
 
 def import_metadata(locationOfMetadata):
     try:
@@ -77,17 +80,18 @@ def generate_knowledge_graph(ontologyPath, metadataPath):
 
     #I need to redefined the property because the property was not defined as functional
     with impO:
-        infoBearing = cco.InformationBearingEntity
-        class has_decimal_value(DataProperty, FunctionalProperty): # Redefining that shit as functional
-            domain    = [cco.InformationBearingEntity]
-            range     = [float]
+        # infoBearing = cco.InformationBearingEntity
+        # class has_decimal_value(DataProperty):
+        #     domain    = [cco.InformationBearingEntity]
+        #     range     = [float]
         class has_text_value(DataProperty, FunctionalProperty):
             domain    = [cco.InformationBearingEntity]
             range     = [str]
-    
+        print(xml)
     #Extract metadata
     metadata = import_metadata(metadataPath)
-    
+    rawPath = os.path.join(baseDir, 'usecases', 'MinimumWorkingExample', 'Data', 'E-modul',
+            metadata['experimentName'])
     ###########################################################################
     ########################ADD INDIVIDUALS####################################
     ###########################################################################
@@ -134,13 +138,13 @@ def generate_knowledge_graph(ontologyPath, metadataPath):
     #Add path to diameter value
     concreteSpecimenIndividual.RO_0000086 = [specimenDiameter]
     specimenDiameter.RO_0010001 = [specimenDiameterValue]
-    specimenDiameterValue.has_decimal_value = metadata['diameter']
+    specimenDiameterValue.has_decimal_value = [float(metadata['diameter'])]
     specimenDiameterValue.uses_measurement_unit = [cco.MillimeterMeasurementUnit]
    
     #Add path to length value
     concreteSpecimenIndividual.RO_0000086 =[specimenLength]
     specimenLength.RO_0010001 = [specimenLengthValue]
-    specimenLengthValue.has_decimal_value = metadata['length']
+    specimenLengthValue.has_decimal_value = [float(metadata['length'])]
     #Add length unit
     if(metadata['length_unit'] == "mm"):
         specimenLengthValue.uses_measurement_unit = [cco.MillimeterMeasurementUnit]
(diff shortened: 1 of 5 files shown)


tests/data_extraction/test_metadata_extraction_emodul.py
```diff
@@ -7,7 +7,7 @@ def test_metadata_extraction_emodul():
     """Tesing the metadata extraction on a single example"""
 
     # setting up the test example
-    input = '../usecases/MinimumWorkingExample/Data/E-modul/BA-Losert MI E-Modul 28d v. 04.08.14 Probe 4'
+    input = '../../usecases/MinimumWorkingExample/Data/E-modul/BA-Losert MI E-Modul 28d v. 04.08.14 Probe 4'
```
eriktamsen, line 10: This change needs to be reversed.
As specified within: [lebedigital.yml](https://github.com/BAMresearch/LebeDigital/blob/main/.github/workflows/lebedigital.yml), the actions run pytest within the tests folder. If you look at the main branch you can see it working.

eriktamsen (changes requested):
This is starting to look good.
I have two more general comments.
1) The test_mapping_script tests both the generation of the graph as well as the query of the data. Is is possible (useful) to separate this?
2) Just now, I did not see how or where the processed data is integrated. This is very important. Maybe for now it might be sufficient to add another path, just like the "rawFileName" but I assume we actually want more, maybe a list with the keys to the dictionary or something to know what is in the processed data.

lebedigital/mapping/emodul_mapping.py
```diff
@@ -1,495 +1,189 @@
...
+            metaDataDict = yaml.load(metaDataFile, Loader=SafeLoader)
+    except Exception as e:
+        print("Path error: " + locationOfMetadata + "do not exist.", file=sys.stderr)
+        print(e, file=sys.stderr)
+    return metaDataDict
+
+def import_ontology(locationOfOntology,my_world):
+    assert os.path.isfile(locationOfOntology), f"Could not find '{locationOfOntology}'"
+    imp_ontology = my_world.get_ontology(locationOfOntology).load()
+    return imp_ontology
```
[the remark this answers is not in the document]
PoNeYvIf: I just got rid of the whole function because after second thought it was not necessary.

tests/data_extraction/test_mapping_script.py
```diff
@@ -0,0 +1,115 @@
...
+    g = rdflib.Graph()
+    result = g.parse(filename, format='ttl')
+    print("Query knowledgeGraph:")
+    
+    print_first()
+
+    result = query_and_test_result('diameter','ns1:informationbearingentity1', 
+                        'ns7:has_decimal_value', g)
+    for row in result:
+        assert target_data['diameter'] == float(row.o)
```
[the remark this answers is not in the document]
PoNeYvIf: It's true that in our metadata for the moment we have only one value, hence, the looping seem useless. But the thing is that the result of the query could return a list with more than one element. In this case the loop will go through and probably fault. It's just a way to make sure that I don't add anything else as a property of diameter.

eriktamsen (approved):
Looks fine to me as a first step.
We will merge and try to apply it in the calibration example and go from there.

tests/data_extraction/test_mapping_script.py
```diff
@@ -0,0 +1,117 @@
...
+    print("Generate knowledgeGraph:")
+    generate_knowledge_graph(metadataPath, filename)
+    
+    g = rdflib.Graph()
+    result = g.parse(filename, format='ttl')
+    print("Query knowledgeGraph:")
+    
+    print_first()
+
+    result = query_and_test_result('diameter','ns1:informationbearingentity1',
```
joergfunger, line 60: Wouldn't it be much more important to query the diameter for a given experiment (thus traversing the knowledge graph) instead of checking a single triple?

tests/data_extraction/test_mapping_script.py
```diff
@@ -0,0 +1,117 @@
...
+    print_first()
+
+    result = query_and_test_result('diameter','ns1:informationbearingentity1', 
+                        'ns7:has_decimal_value', g)
+    for row in result:
+        assert target_data['diameter'] == float(row.o)
+    
+    print_next()
+    
+    result = query_and_test_result('length', 'ns1:informationbearingentity2',
```
joergfunger, line 67: same here, I'm not sure if it helps to "wrap" the query into another query_and_test function. I understand that it might make the testing easier to write, but I guess this is not how people are using the knowledgegraph at the end (in particular if more than a single test is added). So I would suggest really adding the query here.

tests/data_extraction/test_mapping_script.py
```diff
@@ -0,0 +1,115 @@
...
+    for row in result:
+        assert target_data['weight'] == float(row.o)
+    
+    print_next()
+    
+    time = get_date_time_value(target_data)
+    result = query_and_test_result('operator_time', 'ns1:informationbearingentity6',
+                        'ns1:has_datetime_value', g)
+    for row in result:
+       assert time == datetime.datetime.strptime(str(row.o).replace('T',' '), "%Y-%m-%d %H:%M:%S")
```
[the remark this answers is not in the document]
joergfunger: Same question, what is the date format that we use in the ontology? Is the return value always a string, or can we include the conversion routine somewhere into the query or as a general standard (e.g. including date/time, float). In particular, we have to make sure that the format is stored in the ontology (such that the problem of which conversion e.g. to float or date/time) is automated and not depending on each individual extraction script.

usecases/MinimumWorkingExample/dodo.py
```diff
@@ -57,3 +61,19 @@ def task_extract_processed_data_emodul():
...
+    for f in os.scandir(metadata_emodulus_directory):
+        if f.is_file():
+            metadata_file_path = Path(f.path)
+            name_of_ttl = f.name.replace('.yaml', '.tll')
+            knowledge_graph_file = Path(knowledge_graphs_directory,
+                                        name_of_ttl)
+            yield{
+                'name': knowledge_graph_file,
+                'actions': [(generate_knowledge_graph, [metadata_file_path,
+                                                    name_of_ttl])],
```
joergfunger, line 76: This generates a knowledge graph for each individual experimental test, but at the end we would like to have one single knowledgegraph for all tests used in a query. The idea was to perform a SHACL validation for each subgraph and then put them together.

joergfunger (commented):
Thanks for the test scripts. There are several remarks to be discussed before merging, in particular I would suggest to actually include the queries and start from the test id (instead of querying only single triples). I would also suggest of including the conversion based on the data type stored in the ontology, instead of manually doing that for each data type. Finally, we have to joint the knowledge graphs, though we could actually do that in a later stage (in order to do the SHACL validation in between).

tests/data_extraction/test_mapping_script.py
PoNeYvIf: I don't really know how to traverse the knowledge graph using SPARQL queries but I will get in touch with Andre and try to apply his suggestion in my tests. I may be interesting to discuss about a standard nomenclature that I could use for getting rid of names like informationbearingentity1 that are generated automatically by owlready. Maybe something like nameOfExperimentDiameter or nameOfSpecimentDiameter.

tests/data_extraction/test_mapping_script.py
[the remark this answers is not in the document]
PoNeYvIf: As far as the date format is concerned the problem comes from the fact that owlready store it this way year-month-dayThh:mm:ss. I tried to change the default behavior but I didn't find a way to change it. The type is xsd:dateTime and the valid format is basically the one used by owlready so in theory every date will be returned this way.

usecases/MinimumWorkingExample/dodo.py
PoNeYvIf: I had a misunderstanding of the task apparently. I thought that the SCHACL validation and the merge of the KGs would be done by another script. I will try to do that step but I don't really know what is the SHACL validation so I would find it helpful if someone could explain it to me or just pointed out a documentation about it.

joergfunger: See my comment in #48 , we will merge this, and afterwards join the knowledge graphs. The important thing is though that all the queries work on the joint graph, not on graphs for individual single experiments.

tests/data_extraction/test_mapping_script.py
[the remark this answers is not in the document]
joergfunger: Not sure if I understand that, where does owlready specify the format? You can easily just store a string that has any format you would like to have?

tests/data_extraction/test_mapping_script.py
joergfunger: Getting rid of informationbearingentity1  is tricky since this is part of the class structure prescribed by the ontology (not by owlready), i.e. this is something you should discuss with @StephanPirskawetz  and @firmao together. As for the complete sparql query, please talk to @firmao .

PR #45, opened 2022-08-29
Debug dodo.py for MinimumWorking example for KG generation

PoNeYvIf:
There was a small bug in the dodo.py in the MinimalWorkingExample. I add a condition so that the .gitignore is not mistaken as a yaml file and stored the KGs in the right directory.

### DIFF
--- a/usecases/MinimumWorkingExample/dodo.py
+++ b/usecases/MinimumWorkingExample/dodo.py
@@ -67,7 +67,7 @@ def task_export_knowledgeGraph_emodul():
     for f in os.scandir(metadata_emodulus_directory):
         if f.is_file() and f.name != ".gitignore":
             metadata_file_path = Path(f.path)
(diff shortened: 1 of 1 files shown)


usecases/MinimumWorkingExample/dodo.py
```diff
@@ -65,15 +65,15 @@ def task_extract_processed_data_emodul():
 #generate knowledgeGraphs
 def task_export_knowledgeGraph_emodul():
     for f in os.scandir(metadata_emodulus_directory):
-        if f.is_file():
+        if f.is_file() and f.name != ".gitignore":
             metadata_file_path = Path(f.path)
             name_of_ttl = f.name.replace('.yaml', '.tll')
```
eriktamsen, line 70: There is a mistake here as well.
The file formate is ttl not tll

eriktamsen (changes requested):

eriktamsen (approved):

PR #53, opened 2022-09-13
Unittest mixture metadata extraction

soudehMasoudian:
a unittest has been written for mixture_metadata_extraction script.


lebedigital/raw_data_processing/metadata_extraction/mixture_metadata_extraction.py
```diff
@@ -0,0 +1,144 @@
...
+                    "Antrags-/Projektnummer": exceltodf.iloc[3,3],
+                    "Bezeichnung der Proben" : exceltodf.iloc[4,3], 
+                    "Betonsorte u -festigkeitsklasse" : exceltodf.iloc[6,4],
+                    "Wasserzementwert" : exceltodf.iloc[7,2],
+                    "Konsistenz" : exceltodf.iloc[7,7],
+                    "Sieblinie n. DIN 1045" : exceltodf.iloc[8,2],
+                    "Körnungsziffer" : exceltodf.iloc[8,8],
+                    "Vol Leim/Zuschlag" : exceltodf.iloc[10,10]
+                    }
+            word = "Zusatzstoff"
```
eriktamsen, line 111: Why do we need this, if we have a "translate" function? Cant we add these pairs there as well?

lebedigital/raw_data_processing/metadata_extraction/mixture_metadata_extraction.py
```diff
@@ -0,0 +1,144 @@
...
+        specimeninfo = pd.Series(specimeninfo)
+        specimeninfo = specimeninfo.fillna("---") # fill nans
```
eriktamsen, line 113: Adding None to the fields might be better then a "---" string.

tests/data_extraction/test_mixture_metadata_extraction.py
```diff
@@ -0,0 +1,55 @@
...
+
+    # setting up the test example
+    input = '../usecases/MinimumWorkingExample/Data/Mischungen/2014_08_05 Rezeptur_MIII.xlsx'
+    target_data = {'Cement -- Quantity in mix': 330, 'Cement -- Specific gravity': 3.123,
+                   'Cement -- Annotation': 'CEM I 42,5 R', 'Water (total) -- Quantity in mix':
+                    175, 'Water (total) -- Specific gravity': 1, 'Water (total) -- Annotation': 
+                    '---', 'Water (effective) -- Quantity in mix': 
+                    '---', 'Water (effective) -- Specific gravity': 
+                    '---', 'Water (effective) -- Annotation': '---', 'Air -- Quantity in mix': 
+                    '---', 'Air -- Specific gravity': '---', 'Air -- Annotation': '---',
```
eriktamsen, line 17: There seem to be values missing.
Where is "Cement -- Quantity in mix" or "Water (total) -- Specific gravity"?

tests/data_extraction/test_mixture_metadata_extraction.py
```diff
@@ -0,0 +1,55 @@
...
+                    '0,1 / 0,5 -- Annotation': 'Okrilla  ', '0,5 / 1,0 -- Quantity in mix': 194, 
+                    '0,5 / 1,0 -- Specific gravity': 2.63,
+                    '0,5 / 1,0 -- Annotation': 'Okrilla  ', '1,0 / 2,0 -- Quantity in mix': 194, 
+                    '1,0 / 2,0 -- Specific gravity': 2.63, '1,0 / 2,0 -- Annotation': 'Okrilla  ', 
+                    '2,0 / 4,0 -- Quantity in mix': 126, '2,0 / 4,0 -- Specific gravity': 2.63, 
+                    '2,0 / 4,0 -- Annotation': 'Okrilla  ', '4,0 / 8,0 -- Quantity in mix': 126, 
+                    '4,0 / 8,0 -- Specific gravity': 2.63, '4,0 / 8,0 -- Annotation': 'Okrilla  ', 
+                    '8,0 / 16,0 -- Quantity in mix': 708, '8,0 / 16,0 -- Specific gravity': 2.63, 
+                    '8,0 / 16,0 -- Annotation': 'Okrilla  ', 'nan -- Quantity in mix': '---', 
+                    'nan -- Specific gravity': '---', 'nan -- Annotation': '---',
```
eriktamsen, line 39: These empty lines should not be included in the data.

eriktamsen (changes requested):
This seems to be a good step, but some imporvements will be needed before merging.

First of all, please make sure the tests are working within your branch. It might be a good idea to pull from main as some things have changed, maybe this will solve the problems in actions.

I would suggest to change the folder strucutre slightly in the "raw_data_processing" directory. A folder "mixture_data" for this script. I will create a new folder (in a new branch) "emodul_data" where the "emodul_metadata_extraction.py" and the "emodul_generate_processed_data.py" will be moved.

Apart from some of the comments in the code, I think we (@joergfunger: people wo know concrete and the people wo need to work with the data) need to think of a better target structure of the data. Probaly do a similar split as with the experimental files in "metadata - yaml file" and "processed data - csv table".

lebedigital/raw_data_processing/metadata_extraction/mixture_metadata_extraction.py
```diff
@@ -0,0 +1,144 @@
...
+
+        Parameter
+        ---------
+        df
+            The dataframe with the german labels that should be translated.
+        dic
+            The excelfile that contains the translations, columns are "german" and "english"
+    """
+    
+    translation = pd.read_excel(dic)
```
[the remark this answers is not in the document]
joergfunger: I think this is to simplify the reading in of the raw excel sheet (which cannot be changed) and to translate the file. However, I'm not sure why we need the translation of the orginal file. IMO, all the raw excel sheets are German, and I think it would be good to a have mapping such that when printing something, we could always change the printed output by a dictionary e.g. `translate[German,specimen]=Probekörper`, but internally I would only use a single (English) representation. That makes it much easier to work with that.

lebedigital/raw_data_processing/metadata_extraction/mixture_metadata_extraction.py
joergfunger: see above, I agree that I would not switch the language of the metadata.yaml, but rather have a translate function to could convert the meta.yaml to meta_ge.yaml, though I'm not sure why we would need that.

lebedigital/raw_data_processing/metadata_extraction/mixture_metadata_extraction.py
joergfunger: Why do we want to add nan at all, I would just leave out the coresponding entry. IMO there is no huge difference between checking if a value exists or whether it is NaN or ---. By leaving it out, we also allow to condense the information, since some informations might be only given for a small subset of the data.

lebedigital/raw_data_processing/metadata_extraction/mixture_metadata_extraction.py
```diff
@@ -0,0 +1,144 @@
...
+            word = "Zusatzstoff"
+        specimeninfo = pd.Series(specimeninfo)
+        specimeninfo = specimeninfo.fillna("---") # fill nans
+        exceltodf.rename(columns=exceltodf.iloc[17], inplace = True) # set column header
+        exceltodf["Stoffart"] = exceltodf["Stoffart"].str.strip()  # remove whitespace
+        exceltodf["Sonstiges/Bemerkungen"] = exceltodf["Sonstiges/Bemerkungen"].str.strip()  # remove whitespace
+        
+        # Some files have multiple entries of Zusatzstoffe, which are not further labeled, so this adds 
+        # the type of Zusatzstoff
+        howmanyzusatzstoffe = [True if i == word else False for i in exceltodf["Stoffart"] ]
```
joergfunger, line 120: Could we have the naming in the code with a consistent language?

joergfunger, line 120: See above, my suggestion is to use a dict such as 
`exceltopdf[translate[language]['type']]` with language being set to german if the read in file is a German raw data file.

tests/data_extraction/test_mixture_metadata_extraction.py
```diff
@@ -0,0 +1,55 @@
...
+    # setting up the test example
+    input = '../usecases/MinimumWorkingExample/Data/Mischungen/2014_08_05 Rezeptur_MIII.xlsx'
+    target_data = {'Cement -- Quantity in mix': 330, 'Cement -- Specific gravity': 3.123,
+                   'Cement -- Annotation': 'CEM I 42,5 R', 'Water (total) -- Quantity in mix':
+                    175, 'Water (total) -- Specific gravity': 1, 'Water (total) -- Annotation': 
+                    '---', 'Water (effective) -- Quantity in mix': 
+                    '---', 'Water (effective) -- Specific gravity': 
+                    '---', 'Water (effective) -- Annotation': '---', 'Air -- Quantity in mix': 
+                    '---', 'Air -- Specific gravity': '---', 'Air -- Annotation': '---', 
+                    'Supplementary cementious materials Kalksteinmehl -- Quantity in mix': 147,
```
joergfunger, line 18: Kalksteinmehl is German, could we make the yaml file language consistent?

tests/data_extraction/test_mixture_metadata_extraction.py
```diff
@@ -0,0 +1,55 @@
+import pytest
+from lebedigital.raw_data_processing.metadata_extraction \
```
joergfunger, line 2: adopt the suggested changes in main from @eriktamsen in #55

joergfunger (changes requested):
Thanks for the nice work. There are a couple of suggested changes. In addition, add the extraction not only to the test, but also to the doit script such that all the mixes are extracted for the minimum working example. In addition, make sure that the github actions do not fail, otherwise a merge is not possible.

lebedigital/raw_data_processing/metadata_extraction/mixture_metadata_extraction.py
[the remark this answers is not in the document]
eriktamsen: No, this specifically is for the translation file: [translation.xlsx]( https://github.com/BAMresearch/LebeDigital/blob/unittest-mixture-metadata-extraction/lebedigital/raw_data_processing/metadata_extraction/translation.xlsx). If we want/need a translation file, I would prefer it to be not an exel file.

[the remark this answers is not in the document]
joergfunger: Yes, you are right, that is correct, my mistake.

joergfunger:
@soudehMasoudian What is the status with this merge request? The mixture meta data extraction is merged, but a test is missing. In addition, it would be good to reflect the structure in lebedigital also in the tests (instead of having a folder in tests that is called data extraction)

soudehMasoudian:
> @soudehMasoudian What is the status with this merge request? The mixture meta data extraction is merged, but a test is missing. In addition, it would be good to reflect the structure in lebedigital also in the tests (instead of having a folder in tests that is called data extraction)

I commited _mixture_metadata_extraction_ and other unittests located inside data_extraction. The tests didn't run properly as an extra package added to tests package. In this case, the pytest doesn't work and exceptions occures. I've changed some parts of unittests and added a class definition with unittest as the input parameter. 
Everything works now, but please give me more specific data if you want to change the packages. Do you want to change the tests package to `raw_data_processing>mixture` for _Test_Mixture_Metadata_Extraction_, `mapping>emodul_mapping` for _test_mapping_script_ and etc?

eriktamsen:
@soudehMasoudian, I dont fully follow what the problem with pytest is/was. 
I dont think mixing two testing packages is a good idea. All our other tests are run in pytest.
I will have a quick look and try to rewrite the function for pytest. Maybe I can understand why there should be a problem.
@joergfunger: I agree with the idea, that the tests and lebedigital directory should have the same structure. Fortunately changing the location of the test scripts is a simple thing. We should maybe have a discussion at some point, if the lebedigital directory could have a better structure, but this is not the place ;)

eriktamsen:
Actually I noticed, that we have at least one more other test done with unittest. I am not sure if these tests are currently automatically included in the CI pytest. Maybe it is not a problem. I would still prefer to stick to one package.

soudehMasoudian:
> I restructured the tests directory, not only for the extraction scripts. after merging, I will make sure all current test scripts in main are in the right place. I rewrote the function for pytest and I guess the problem was the "nan" value, however I checked and unittest tests are working fine in combination with the pytest package. we can use both packages if we want to and do not necessarily need to change the others.

If you're speaking about lebedigital scripts it's true. Json doesn't know nan, rather it should be camel case, like _NaN_. Except from that, the array keys should be sorounded with double quotation insterad of single to be readable as a json for test. Besides that, it's much better to compare the equality of two json than strings as it's more accurate and less error prone. But it works perfectly when we define the service result as a string.  
Now I'll change the structure of the unittest packaging once you confirm it.

tests/mapping/test_mapping_script.py
```diff
@@ -0,0 +1,121 @@
+import unittest
+import rdflib
+import datetime
+import os
+import yaml
+from pathlib import Path
+
+from lebedigital.mapping.emodul_mapping import generate_knowledge_graph, get_date_time_value
+
+class TestMappingScript(unittest.TestCase):
```
joergfunger, line 10: mapping is done for different experiments, in that case Youngs modulus test (which is called emodul_mapping). Thus, I would suggest to have a file in test/mapping that is called test_emodul_mapping.py). I would also suggest we stick to a single testing framework  (as Erik has already pointed out) and use pytest instead of unittest.

tests/mapping/test_mapping_script.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def query_and_test_result(self, typeOfQuery, predicate, prop, g):
+        print(f"""Query for {typeOfQuery}:""")
```
joergfunger, line 13: It would be advantageous to not use print but rather the loguru logger such that we can define (when running the tests) the level of output we are interested in.

tests/mapping/test_mapping_script.py
```diff
@@ -0,0 +1,121 @@
...
+                       'specimen_name': 'BA-Losert E-Modul 28d v. 04.08.14 Probe 4',
+                       'remark': 'Kraftgeregelt 3,9 kN/s',
+                       'weight': 5342.0,
+                       'weight_unit': 'g',
+                       'diameter': 98.6,
+                       'length': 300.3,
+                       'length_unit': 'mm',
+                       'mix_file': '2014_08_05 Rezeptur_MI.xlsx'}
+
+        with open(metadataPath, 'w') as yamlFile:
```
joergfunger, line 53: are these newly created files added to the gitignore

tests/mapping/test_mapping_script.py
```diff
@@ -0,0 +1,121 @@
...
+        print("Generate knowledgeGraph:")
+        generate_knowledge_graph(metadataPath, filename)
+
+        g = rdflib.Graph()
+        result = g.parse(filename, format='ttl')
+        print("Query knowledgeGraph:")
+
+        self.print_first()
+
+        result = self.query_and_test_result('diameter', 'ns1:informationbearingentity1',
```
joergfunger, line 64: This only checks of there is a single informationbearingentity1 with the value. How do we know that this is the diameter? I would think it might be relevant to start from the id of the test and then extract its diameter traversing the complete graph. This might look redundant (and does not allow to have a query_and_test method) but IMO advantageous (since this is how we are supposed to use the knowledge graph afterwards). This is true for all the subsequent tests.

joergfunger (changes requested):
Thanks for the revision. I would stick to a single testing framework (pytest). In addition, I have some comments to the query tests.

eriktamsen:
@joergfunger: I would suggest to open a new issue to improve the query test and refactor them to use pytest only.
These files where not changed (as far as I know) by @soudehMasoudian now. It only looks like it, as I have moved them to a new place.

eriktamsen:
I copied your comments to a new issue: https://github.com/BAMresearch/LebeDigital/issues/84
I hope to assign someone to the issue in the next knowledge graph meeting.
I would suggest to merge this PR.

joergfunger (approved):

tests/mapping/test_mapping_script.py
soudehMasoudian: These files (`testMetadata.yaml`, `knowledgeGraph.ttl` will be removed at the end of the script and  no track of them can be found in the end. So there is no need to add them to gitignore.

tests/mapping/test_mapping_script.py
soudehMasoudian: Could you please give me more info related this issue? How the function should change in order to cover the new requirements? It seems a bit confusing, do you mean it's needed to change the query?

joergfunger: yes, the query is supposed to search in a complete triple store with all test (potentially multiple test each having a diameter)

eriktamsen: @joergfunger are you sure? I thought this test is supposed to check the mapping of extracted data to ttl file.
The connection to the triple store is one or two steps further, right?
This is still an open issue but not related to this test.

joergfunger: The data structure that is generated is a graph. We are currently just checking the last leaf of the graph, without any understanding of the connections. IMO, the test should traverse the complete graph, such that this graph is tested, not only the final leaf. Thus, the question to be answered would be to return the diameter of a Young's modulus test (with a given id) by traversing from the test to the specimen and then the diameter.

PR #55, opened 2022-09-14
restructured raw data processing dir

eriktamsen:
Created a "youngs_modulus_data" directory for the respective raw data processing scripts.
Changed required paths in dodo and test files.


joergfunger (approved):
Good proposal, that is much easier to structure.

PR #57, opened 2022-09-21
44 getting data from kg for calibration

atulag0711:
- Added tests for calibration module.
- Added the calibration module in the dodo script for minimum working example.


lebedigital/calibration/calibrationWorkflow.py
```diff
@@ -0,0 +1,202 @@
...
+        skip_init = 330
+        exp_output = load_experimental_data(experiment_name, skip_init, skip_last, path=path_csv)
+
+    # This generated Latex errors!!! Maybe we require some extra package
+    #plt.show()
+    # ========================================
+    #       Set Numerical values
+    # ========================================
+    # "uninformed" Normal prior for the E modulus
+    loc_E = 100  # kN/mm2
```
joergfunger, line 58: How did you get that prior? Shouldn't that be part of the interface?

lebedigital/calibration/calibrationWorkflow.py
```diff
@@ -0,0 +1,202 @@
...
+    # ============================================
+
+    y_test = exp_output['stress'] * (np.pi * (exp_output['diameter'] / 2) ** 2)  # in kN
+    # add the experimental data
+    problem.add_experiment(
+        experiment_name,
+        fwd_model_name="linear_elasticity_cylinder",
+        sensor_values={
+            'nu': 0.2,
+            'height': exp_output['height'],
```
joergfunger, line 100: just to make sure, what is the height that is expected to be used here (maybe that is a question to @eriktamsen). Is that the height of the specimen, or the measurement length for the displacement sensor?

lebedigital/calibration/calibrationWorkflow.py
```diff
@@ -0,0 +1,202 @@
...
+            'radius': exp_output['diameter'] / 2,
+            'displacement_list': exp_output['displacement'],
```
joergfunger, line 102: same here, what is that displacement (the actual deformation of the specimen (30cm), or the relative displacement of the central part (10cm)

lebedigital/calibration/calibrationWorkflow.py
```diff
@@ -0,0 +1,202 @@
...
+from probeye.inference.emcee.solver import EmceeSolver
+
+# local imports (others)
+from lebedigital.Prediction.three_point_bending_example import three_point_bending_example
+from lebedigital.calibration.utils import query_KG, PosteriorPredictive, load_experimental_data
+from lebedigital.calibration.forwardmodel_linear_elastic_cylinder import LinearElasticityCylinder
+
+
+def perform_calibration(path_to_KG : str, calibrated_data_path : str, experiment_name='Wolf 8.2 Probe 1', mode = 'full', KnowledgeGraph = True):
+    # =========================================
```
joergfunger, line 28: The output of the method should also be documented.

lebedigital/calibration/calibrationWorkflow.py
```diff
@@ -0,0 +1,202 @@
...
+    # ==============================================
+    #        Export knowledge graph
+    # ==============================================
+
+    # create the knowledge graph and print it to file
+    #dir_path = os.path.dirname(__file__)
+    dir_path = calibrated_data_path
+    basename_owl = os.path.basename(__file__).split(".")[0] + ".owl"
+    knowledge_graph_file = os.path.join(dir_path, basename_owl)
+    export_knowledge_graph(problem, knowledge_graph_file, data_dir=dir_path)
```
joergfunger, line 133: Where is that stored (this is hardcoded and for multiple data sets this is overwritten. I would also assume that it might be advantageous to return the graph instead of the file, such that writing to a file could actually be done outside in the doit script rather than the library (or you would have to give the file location as input).

lebedigital/calibration/calibrationWorkflow.py
```diff
@@ -0,0 +1,202 @@
...
+        inference_data = emcee_solver.run_mcmc(
+            n_walkers=4,
+            n_steps=1,
+            n_initial_steps=4,
+        )
+    else:
+        inference_data = emcee_solver.run_mcmc(
+            n_walkers=5,
+            n_steps=40,
+            n_initial_steps=5,
```
joergfunger, line 152: Does this setup result in an accurate estimate?

lebedigital/calibration/calibrationWorkflow.py
```diff
@@ -0,0 +1,202 @@
...
+
+    # ========================================================================
+    #       Posterior Predictive
+    # ========================================================================
+    nu = 0.2
+    E_pos = sample_dict['E'] * 1000  # N/mm2 ~ E_mean ~ 95E03N/mm2
+    # three_point = three_point_bending_example()
+    pos_pred = PosteriorPredictive(three_point_bending_example, known_input_solver=nu, parameter=E_pos)
+
+    if mode == 'cheap':
```
joergfunger, line 178: what is the return value of this method?

joergfunger: Is there a test for this functionality? (so where we know what we would expect?)

lebedigital/calibration/calibrationWorkflow.py
```diff
@@ -0,0 +1,202 @@
...
+    plt.figure()
+    sns.kdeplot(posterior_pred_samples)
+    plt.xlabel('stress in x-direction')
+    # more problems with latex when running doit
+    #plt.show()
+    #plt.savefig(calibrated_data_path + 'posterior_predictive.pdf',dpi =100)
+    # plt.savefig('Figures/posterior_predictive.pdf', dpi=100)
+
+
+# testing things
```
joergfunger, line 194: Can we remove the comments?

lebedigital/calibration/forwardmodel_linear_elastic_cylinder.py
```diff
@@ -23,7 +22,7 @@ def definition(self):
                               Sensor("height"),
                               Sensor("radius"),
                               Sensor("displacement_list")]
-        self.output_sensors = [Sensor('force_list')]
+        self.output_sensors = [Sensor('force_list', std_model="sigma")]
```
joergfunger, line 25: what is std_model? @eriktamsen what is the height here (30cm or 10cm)?

lebedigital/calibration/forwardmodel_linear_elastic_cylinder.py
```diff
@@ -50,21 +49,26 @@ def response(self, inp: dict) -> dict:
         parameters['log_level'] = 'WARNING'
         parameters['bc_setting'] = 'fixed'
         parameters['dim'] = 3
+
+        # as we know this problem is linear elastic, there is no point in solving it multiple times
+        # a test load is applied and then interpolated to the load list
+        test_load = -0.05
```
joergfunger, line 55: Could we somehow store this result in an intermediate yaml file such that for same inputs/outputs we could store the slope (with reading in the file instead of solving might be much faster)

lebedigital/calibration/utils.py
```diff
@@ -0,0 +1,195 @@
...
+    df = pd.read_csv(file_path, skipfooter=skip_last, engine='python')
+    df = df.drop(labels=range(0, skip_init), axis=0)
+    df['displacement'] = (df['Transducer 1[mm]'] + df['Transducer 2[mm]'] + df['Transducer 3[mm]']) / 3
+    df['stress'] = df['Force [kN]'] / (np.pi * (float(results['diameter']) / 2) ** 2)
+
+    output = {
+        'height': float(results['length']),
+        'diameter': float(results['diameter']),
+        'displacement': np.array(df['displacement']),
+        'stress': np.array(df['stress'])}
```
joergfunger, line 82: Why are we passing stress (as already some processed value) instead of the measured force?

lebedigital/calibration/utils.py
```diff
@@ -0,0 +1,195 @@
...
+    return output
+
+def load_experimental_data(exp_name, skip_init, skip_last, KG=False, path=None):
```
joergfunger, line 85: Should we have the scripts that perform some analysis of the data for a specific test go into that directory instead of putting that into calibration?

lebedigital/calibration/utils.py
```diff
@@ -0,0 +1,195 @@
...
+
+        """
+        # Monte carlo step
+        output = []
+        for i in range(0, samples):
+            y = self._forward_solver(self._parameter[i], self._known_input)
+            output.append(y)
+
+        # get the posterior pred stats
+        mean = np.mean(output, axis=0)
```
joergfunger, line 185: You estimate the noise as well, the posterior predictive would require an integration over the latent parameters.

tests/emodulus_calibration/Wolf 8.2 Probe 1.csv
```diff
@@ -0,0 +1,529 @@
+Force [kN],Transducer 1[mm],Transducer 2[mm],Transducer 3[mm]
```
joergfunger, line 1: why is that csv file stored?

usecases/Concrete/Calibration/E_modul_calibration.py
```diff
@@ -1,7 +1,7 @@
 # -- PKM TUM -- atul.agrawal@tum.de -- #
+# 01.02.2022 : Deprecated, please dont use
 
-from misc import load_experimental_data
-import numpy as np
+from lebedigital.calibration.utils import load_experimental_data
```
joergfunger, line 4: should that be removed then?

usecases/MinimumWorkingExample/dodo.py
```diff
@@ -57,7 +60,7 @@ def task_extract_metadata_emodul():
                 'actions': [(emodul_metadata, [raw_data_path, yaml_metadata_file])],
                 'file_dep': [raw_data_file],
                 'targets': [yaml_metadata_file],
-                'clean': [clean_targets]
+                'clean': [clean_targets]  # what does this do?
```
joergfunger, line 63: remove comments

usecases/MinimumWorkingExample/dodo.py
```diff
@@ -120,4 +123,29 @@ def task_export_knowledgeGraph_emodul():
...
+            }
+
+# perform calibration
+def task_perform_calibration():
+    # create folder, if it is not there
+    Path(calibrated_data_directory).mkdir(parents=True, exist_ok=True)
+
+
+
+    # TODO implement "expensive" option on all files!!!
```
joergfunger, line 135: could we have that already here to see if the general implementation is working (even though the CI uses only the cheap version?)

usecases/MinimumWorkingExample/dodo.py
```diff
@@ -120,4 +123,29 @@ def task_export_knowledgeGraph_emodul():
...
+    knowledge_graphs_file = Path(knowledge_graphs_directory, cheap_example_name + '.ttl')
+
+    # current outputs
+    owl_file = Path(calibrated_data_directory, 'calibrationWorkflow.owl')
```
joergfunger, line 139: what is the owl file (the output is not owl but ttl). owl is the ontology, ttl the knowledge graph. In adition, this should be different for each experiment.

usecases/MinimumWorkingExample/dodo.py
```diff
@@ -120,4 +123,29 @@ def task_export_knowledgeGraph_emodul():
...
+    displ_list = Path(calibrated_data_directory, 'displacement_list_' + cheap_example_name + '.dat')
```
joergfunger, line 140: why is the force and displ list an output of the calibration? Where is that defined?

usecases/MinimumWorkingExample/mesh/cylinder.xdmf
```diff
@@ -0,0 +1 @@
+<Xdmf Version="3.0"><Domain><Grid Name="Grid"><Geometry GeometryType="XYZ"><DataItem DataType="Float" Dimensions="158 3" Format="HDF" Precision="8">cylinder.h5:/data0</DataItem></Geometry><Topology TopologyType="Tetrahedron" NumberOfElements="428" NodesPerElement="4"><DataItem DataType="Int" Dimensions="428 4" Format="HDF" Precision="8">cylinder.h5:/data1</DataItem></Topology><Attribute Name="gmsh:dim_tags" AttributeType="Vector" Center="Node"><DataItem DataType="Int" Dimensions="158 2" Format="HDF" Precision="8">cylinder.h5:/data2</DataItem></Attribute><Attribute Name="gmsh:physical" AttributeType="Scalar" Center="Cell"><DataItem DataType="Int" Dimensions="428" Format="HDF" Precision="8">cylinder.h5:/data3</DataItem></Attribute><Attribute Name="gmsh:geometrical" AttributeType="Scalar" Center="Cell"><DataItem DataType="Int" Dimensions="428" Format="HDF" Precision="8">cylinder.h5:/data4</DataItem></Attribute></Grid></Domain></Xdmf>
```
joergfunger, line 1: do we have to store all the binaries in the repo, or could those be generated on the fly?

joergfunger (changes requested):
Thanks for the merge request. Please find attached my comments for the specific files.

usecases/MinimumWorkingExample/mesh/cylinder.xdmf
eriktamsen: they dont need to be stored longterm, this is just a "byproduct", therefore also not found in the doit dependencies and targets. this is due to the gmsh mesh generation and transformation to fenics mesh.
it is overridden when run twice. we could make sure to delete it afterwards to keep everything "clean"

lebedigital/calibration/calibrationWorkflow.py
eriktamsen: @joergfunger: which method?

lebedigital/calibration/calibrationWorkflow.py
joergfunger: The cylinders are usually 30cm in length/height, but the deformation is just measured in the middle for a gauge length of 10cm (to avoid the influence of the friction at the boundaries to actually get an almost uniform stress state. Not sure if this height represents the complete cylinder (30cm) or just the inner part (10cm)?

lebedigital/calibration/calibrationWorkflow.py
joergfunger: see above, the data is only for the inner 10cm, the total height is 30cm, but the gauge length is 10cm.

lebedigital/calibration/calibrationWorkflow.py
joergfunger: That would mean we should (in addition to the accuracy) add a seed for the random number generator.

lebedigital/calibration/calibrationWorkflow.py
joergfunger: for the inference method. Usually, all the methods in the lebedigital modul should be part of the pytest set, such that errors can be detected.

lebedigital/calibration/forwardmodel_linear_elastic_cylinder.py
joergfunger: see above

usecases/MinimumWorkingExample/mesh/cylinder.xdmf
joergfunger: Could we then just add that in the .gitignore file?

eriktamsen: Done!

.github/workflows/lebedigital.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         shell: bash -l {0}
         run: |
           cd $GITHUB_WORKSPACE/tests/
-          pytest -s -W error::UserWarning --ignore=ModelCalibration/tests/simulation
+          pytest -s --ignore=ModelCalibration/tests/simulation
```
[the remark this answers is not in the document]
atulag0711: Yes, its a blank folder . On it.

lebedigital/calibration/calibrationWorkflow.py
```diff
@@ -0,0 +1,202 @@
...
+    #experiment_name = 'Wolf 8.2 Probe 1'
+    # -- if KG is not there, using custom script
+    # path_csv = '../MinimumWorkingExample/emodul/processed_data/Wolf 8.2 Probe 1.csv'
+    # skip_last = 145
+    # skip_init = 330
+    #experiment_name = 'Wolf 8.2 Probe 1'
+    # exp_output = load_experimental_data(experiment_name, skip_init, skip_last, path=path_csv)
+    # -- if KG is present, comment the above
+    #path_to_KG = '../usecases/MinimumWorkingExample/emodul/knowledge_graphs/'
+    if KnowledgeGraph:
```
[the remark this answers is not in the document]
atulag0711: Yes if KG is querried outside the calibration script, it would be removed.

lebedigital/calibration/utils.py
```diff
@@ -0,0 +1,195 @@
...
+from pathlib import Path
+import sys
+
+baseDir1 = Path(__file__).resolve().parents[1]
+sys.path.append(os.path.join(os.path.join(baseDir1, 'knowledgeGraph'), 'emodul'))
+sys.path.append(os.path.join(baseDir1, 'Data'))
+
+#import emodul_query
+
+def query_KG(path: str, skip_last = 145, skip_init = 330 ) -> dict:
```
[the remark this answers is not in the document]
atulag0711: We already discussed this, we are essentially choosing the third loading is all. So it I think skipping the last unloading.

lebedigital/calibration/calibrationWorkflow.py
```diff
@@ -0,0 +1,202 @@
...
+        path = path_to_KG + '/' + experiment_name + '.ttl'
+        exp_output = query_KG(path=path)
+        plt.plot(exp_output['stress'], exp_output['displacement'] / exp_output['height'])  # checking loading data
+    else:
+        path_csv = path_to_KG # if KG not present, path takes path of CSV
+        #path_csv = os.path.dirname(Path(__file__)) + '/Wolf 8.2 Probe 1.csv'
+        skip_last = 145
```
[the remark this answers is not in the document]
atulag0711: Response above, yes file specific.

lebedigital/calibration/calibrationWorkflow.py
atulag0711: I didnt understand your question. Its a uninformed prior, see the sd is very high. Setting the value of mean is a bit irrelevant in this case.

lebedigital/calibration/calibrationWorkflow.py
atulag0711: Yes alright.

lebedigital/calibration/calibrationWorkflow.py
atulag0711: Yes the walker and no steps are not enough. Need more. Also seed is needed, atleast for the tests. I will see how I can do so.

lebedigital/calibration/calibrationWorkflow.py
atulag0711: I am not sure I understand this comment. Yes you are right, it generates an .owl file and henceforth the KG for a given dataset, but for multiple dataset. But cant be move forward with this limited functionality for now?

lebedigital/calibration/calibrationWorkflow.py
atulag0711: For teh calibration script there is a test. If you mean for teh posterior predictive here, there is none and can be added

lebedigital/calibration/utils.py
atulag0711: No particular reason. Was easier for me to do calculations downstream.

lebedigital/calibration/utils.py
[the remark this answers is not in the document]
joergfunger: There are two challenges. 145 is valid only for a very specific data set, in general, we don't know how many time steps have been recorded (even for different E-modulus tests, the skip init and skip last could be different and have to be determined by looking at the function itself - automatically detecting the third loading cycle). Furthermore (and that is the same with the above challenge). This is under lebedigital/calibration/utils.py -  where is that path or the naming of that function implying that this refers to the Youngs-modulus data?

lebedigital/calibration/utils.py
atulag0711: Can you please elaborate what is "that" directory. Didnt get it.

lebedigital/calibration/utils.py
atulag0711: I assume the noise to be sensor noise, not model discrepancy, so dont need to include that IMO.

tests/emodulus_calibration/Wolf 8.2 Probe 1.csv
atulag0711: We already discussed this, just for the test as in the dodo it is generated on the fly and this is nor present by default.

usecases/MinimumWorkingExample/dodo.py
atulag0711: ProbEye was just exporting it somehow. I didnt put that much emphasis on that. I will see.

lebedigital/calibration/utils.py
joergfunger: In load_experimental_data you load in a file spefically related to a Young's modulus test (this would not work for the hydration test data). How would you name the load_experimental_data for the hydration tests, since this name is already taken? Also, the calibration that you perform is specifically related only to Young's modulus data (the hydration tests would require different inputs, outputs, models, ..). Thus, the suggestion is to move that to a directory related to the corresponding experiment, e.g. calibration/labtests/YoungsModulus.

environment.yml
```diff
@@ -26,5 +27,5 @@ dependencies:
   - pytest
   - pip
   - pip:
-    - probeye==2.3.2
+    - probeye==2.3.0
```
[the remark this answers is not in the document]
eriktamsen: makes sense. I did not know that. I thought this is supposed to be one of the features of probeye.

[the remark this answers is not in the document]
joergfunger: The current version still has the feature??

lebedigital/calibration/calibrationWorkflow.py
joergfunger: If we loop in our simple usecase over different tests, then there is no way we could keep these different KG? Or attach it to the global KG?

lebedigital/calibration/calibrationWorkflow.py
joergfunger: That depends on the number of data points and then weight you are actually giving. And an uninformed prior should have a clear phyiscal meaning, in your case the normal distribution results in values of the Youngs modulus that can be negative, which is not physical.

joergfunger: In addition, I suggest that this is not hard coded here, but added by the user (potentially in the doit scripts or in the pytests, but not here in the general routines that should work for any Youngs modulus test.

lebedigital/calibration/calibrationWorkflow.py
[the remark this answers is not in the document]
joergfunger: So how would we then analyse all data files in the minimum working example?

lebedigital/calibration/calibrationWorkflow.py
eriktamsen: The simulation has been changed to fixed 10cm "height" and the boundary conditions changed to match the unconfined conditions.

lebedigital/calibration/calibrationWorkflow.py
eriktamsen: The simulation has been changed to fixed 10cm "height" and the boundary conditions changed to match the unconfined conditions.

eriktamsen:
Hey @atulag0711, I tried to get an overview of the status of this branch.
This PR is quit messy. It would be nice if you could go through all comments, resolve the ones that are clear and continue the discussions on the ones that are not clear yet.
The few points I recognized, that definitely require changes when looking through the PR, is 1) automatically identifying the third load step (skip_last and skip_init) 2) giving the files and functions better names.
There are probably some small details, but it will be nice to move on/finish this minimum working example!

atulag0711:
Hey Erik. @eriktamsen Agreed. In view of your comments and the past comments. I updated and pushed. As discussed, since the "calibration function" now is only taking inputs as dict and not calling the KG internally, the correct load-displacement curve (third cycle) is to be provided by the KG team.

eriktamsen:
Thanks for the update. Could you still please resolve the conversations here, so that we know which points are open and which have been addressed.
The actions fail, which might be resolved when merging main. Then there are conflicts with .gitignore. Please resolve them.
Concerning the "third loading", I personally would disagree, but I see that there can be arguments made to move this to a preprocessing step. @joergfunger, what do you think? My main point is a practical one. I think it is easier for you to do this, than communicating and delegating this to some HiWi, most of who are currently on a break due to exams.

joergfunger:
As for the "third loading cycle", I'm not completely sure that this is to be done by the knowledge graph team, since even though this is some kind of preprocessing for the final calibration, I'm not sure if this could/should be done automatically and only this final step should be stored. There is e.g. the question of noise in the data. If you look at it, it is not completely clear how we should define the start and end of the third load cycle, since the minimum (beginning) and maximum (end) of the cycle might not directly be the start and enf of the cycle. Thus, this could be done in several ways, e.g. a) one being to fit a two linear functions (constant phase and linear phase and take the intersection as the beginning b) using AIC (Akaike information criterion)  to detect the changes c) include this start/end as variables in the identification (after some rough preprocessing extracting only the last constant/linear increase)constant part. As such, I would argue that it might be quite difficult to hand that over to some student assistant.

environment.yml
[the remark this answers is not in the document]
atulag0711: I am using the older because it lets me query the KG also. I think ALex told its not there in the new because of some reasons.

tests/emodulus_calibration/test_emodulus_calibration.py
```diff
@@ -0,0 +1,41 @@
...
+        experimental_data=output, calibration_metadata={"E_loc": 100, "E_scale": 100},
+        calibrated_data_path=os.path.dirname(Path(__file__)), mode='cheap'
+    )
+    # checking if the KG file for the calibration is created
+    assert os.path.exists(os.path.join(os.path.dirname(Path(__file__), 'calibrationWorkflowWolf 8.2 Probe 1.csv.owl'))) == True
+
+    # checking for the calibrated E modulus and the posterior predictive
+    assert np.mean(post_pred) == pytest.approx(95, rel=50)
+
+    assert np.mean(E_samples) == pytest.approx(95E03, rel=50e03)
```
joergfunger, line 27: There is something wrong here, the E-modulus of these tests should not be 95e3Mpa. That is too large.

joergfunger (changes requested):

lebedigital/calibration/utils.py
```diff
@@ -0,0 +1,203 @@
...
+    if KG is not True:  # manually passing the processed data path
+        df = pd.read_csv(path, skipfooter=skip_last)
+    df = df.drop(labels=range(0, skip_init), axis=0)
+
+    if KG is True:
+        dia = emodul_query.input_emodul_data_for_calibration(exp_name)['specimenDiameter']
+        height = emodul_query.input_emodul_data_for_calibration(exp_name)['specimenLength']
+    if KG is not True:
+        dia = 98.6
+        height = 300.2
```
eriktamsen, line 118: Height needs to be changed to 100mm as this is the gauge length.

lebedigital/calibration/utils.py
```diff
@@ -0,0 +1,203 @@
...
+
+        # read csv into dataframe
+        file_path = results['path'] + '/' + results['file_name']
+    #skip_last = 145 # hard coded to get the third loading cycle. need somthing better
+    #skip_init = 330
+    if KG == False: # this is just for testing purposes
+        file_path = os.path.join(path,exp_name)
+        results = {}
+        results['length'] = 98.6
+        results['diameter'] = 300.2
```
eriktamsen, line 79: length and height seem to be switched.
Height should 100mm (gauge length), not 300mm (actual specimen length)

tests/emodulus_calibration/test_emodulus_calibration.py
eriktamsen: I think this might be the result of using the specimen length (around 300mm) instead of the gauge length (100mm) in some places.

lebedigital/calibration/utils.py
```diff
@@ -0,0 +1,203 @@
...
+
+    if KG is True:
+        dia = emodul_query.input_emodul_data_for_calibration(exp_name)['specimenDiameter']
+        height = emodul_query.input_emodul_data_for_calibration(exp_name)['specimenLength']
+    if KG is not True:
+        dia = 98.6
+        height = 300.2
+
+    df['displacement'] = (df['Transducer 1[mm]'] + df['Transducer 2[mm]'] + df['Transducer 3[mm]']) / 3
+    df['stress'] = df['Force [kN]'] / (np.pi * (dia / 2) ** 2)
```
eriktamsen, line 121: Can we reduce the if statements to a single if - else instead of 4 if statements?

eriktamsen, line 121: Why is the stress computed? probably leftovers from initial implementation I guess.
The data is given in force, the simulation returns force. This seems to make things more complicated.

lebedigital/calibration/utils.py
atulag0711: This is an old method, not in use as you would have seen.

lebedigital/calibration/utils.py
atulag0711: This is an old method, not in use as you would have seen. I treat the physical problem as black box, so the correct data is expected to be provided by the Knowledge graph folks.

atulag0711: This is an old method, not in use as you would have seen.

lebedigital/calibration/utils.py
atulag0711: Ah you are right. My bad. Thanks for pointing out. Fixed.

tests/emodulus_calibration/test_emodulus_calibration.py
atulag0711: @joergfunger alright, then in what range it should be? I remember asking this multiple time that what is the expected range  then and not getting any input from domain experts/ or what is the standard E modulus from ususal methods (maybe standards).

eriktamsen: Its expected to be around 30000 N/m2 (or Pa). The value is about 3 times too large. Thats, why I think it might have something todo, with applying the wrong length somewhere.

lebedigital/calibration/forwardmodel_linear_elastic_cylinder.py
atulag0711: It was part, necessary for the version of ProbEye, I dont remember why.

lebedigital/calibration/calibrationWorkflow.py
```diff
@@ -0,0 +1,260 @@
...
+    # =======================================================================
+
+    # get the samples from the knowledge graph
+    sample_dict = import_parameter_samples(knowledge_graph_file)
+
+    # ========================================================================
+    #       Posterior Predictive
+    # ========================================================================
+    nu = 0.2
+    E_pos = sample_dict["E"] * 1000  # N/mm2 ~ E_mean ~ 95E03N/mm2
```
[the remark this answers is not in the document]
atulag0711: conversion from kN//mm2 to N/mm2 to accommodate for the units used in three point bending test example.

lebedigital/calibration/utils.py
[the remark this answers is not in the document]
atulag0711: Added a method to extract third loading cycle

eriktamsen (changes requested):
Thanks for providing the extraction of the third load cycle, I was not able to easily check it but it looks good. I added some small comments, which you can see more as suggestions, but the main 3 things are:
- the tests are not working
- there are unresolved conversations
- there are merge conflicts with main

@atulag0711: Please sort this out to be able to finally merge this branch.

lebedigital/calibration/utils.py
```diff
@@ -156,8 +137,8 @@ def __init__(self, forward_solver, known_input_solver, parameter = None, query_k
             needed for the forward model
         """
 
-        #assert (parameter == np.array), "Parameter needs to be np.ndarray"
-        #assert parameter.shape[1] == 1, "Currently support only 1 parameter"
+        # assert (parameter == np.array), "Parameter needs to be np.ndarray"
+        # assert parameter.shape[1] == 1, "Currently support only 1 parameter"
 
         self._forward_solver = forward_solver
         if parameter is not None:
```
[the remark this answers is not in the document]
atulag0711: this is very temporary. will later iron out the sharp edges

lebedigital/calibration/utils.py
```diff
@@ -51,88 +52,68 @@ def query_objects(queries, graph):
...
         # read csv into dataframe
-        file_path = results['path'] + '/' + results['file_name']
-    #skip_last = 145 # hard coded to get the third loading cycle. need somthing better
-    #skip_init = 330
-    if KG == False: # this is just for testing purposes
-        file_path = os.path.join(path,exp_name)
+        file_path = results["path"] + "/" + results["file_name"]
+    # skip_last = 145 # hard coded to get the third loading cycle. need somthing better
+    # skip_init = 330
+    if KG == False:  # this is just for testing purposes
```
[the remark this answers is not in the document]
atulag0711: QUeery is outside only, the above I just delete for just in case. code below line 88 is active.

lebedigital/Prediction/three_point_bending_example.py
```diff
@@ -55,8 +55,10 @@ def three_point_bending_example(E, nu):
 
 # example of how to use this function
 # defining the material parameters
-# E = 30000  # N/mm²
-# nu = 0.2
-# stress = three_point_bending_example(E, nu)
-# # resulting stress in x direction in the bottom center of the beam
-# print(stress)
+#E = 30000  # N/mm² 30072.79961482137
```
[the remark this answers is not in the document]
atulag0711: done

eriktamsen (approved):
Looks okay to me.
Other things can be fixed/improved in new issues

tests/emodulus_calibration/test_emodulus_calibration.py
```diff
@@ -0,0 +1,31 @@
...
+    E_samples= esimate_Youngs_modulus(
+        experimental_data=output, calibration_metadata={"E_loc": 30, "E_scale": 10},
+        calibrated_data_path=os.path.dirname(Path(__file__)), mode='cheap'
+    )
+    # checking if the KG file for the calibration is created
+    assert os.path.exists(os.path.join(os.path.dirname(Path(__file__)),
+                          'calibrationWorkflowWolf 8.2 Probe 1.csv.owl')) is True
+
+    # checking for the calibrated E modulus
+    print(f"The posterior E mean is {np.mean(E_samples)} KN/mm2")
```
joergfunger, line 26: It would be nice if print statements do not clutter our test execution output. If you want to keep them for debugging, there should be an option to turn them off, otherwise the loggers (we use loguru) might also be an option.

tests/emodulus_calibration/test_emodulus_calibration.py
```diff
@@ -0,0 +1,31 @@
...
+import pytest
+import numpy as np
+import os
+from pathlib import Path
+
+def test_emodulus_calibration():
+
+    # query knowledge graph. Note KG team needs to provide this.
+    print("Calibration test started")
+    output = query_KG(
```
joergfunger, line 14: why do you have to query a KG for a simple CSV file? The tests should be unit tests, thus not include other complex tasks. In your case, I would expect the input to be a load displacement curve, the output a posterior of E

tests/emodulus_calibration/test_emodulus_calibration.py
```diff
@@ -0,0 +1,31 @@
...
+    )
+    # checking if the KG file for the calibration is created
+    assert os.path.exists(os.path.join(os.path.dirname(Path(__file__)),
+                          'calibrationWorkflowWolf 8.2 Probe 1.csv.owl')) is True
+
+    # checking for the calibrated E modulus
+    print(f"The posterior E mean is {np.mean(E_samples)} KN/mm2")
+    assert np.mean(E_samples) == pytest.approx(30, rel=0.4) # this needs to specified by subject matter experts
+
+    print("Calibration test completed")
```
joergfunger, line 29: print see above

tests/emodulus_calibration/test_prediction.py
```diff
@@ -0,0 +1,22 @@
...
+import numpy as np
+import os
+from pathlib import Path
+
+
+def test_prediction():
+    print("Prediction test started")
+
+    # check if KG is present
+    kg_path = "prediction_data/prediction_kg.owl"
```
joergfunger: I'm confused here, why do you load an ontology? In addition, the unit test should test the individual function, which here would be IMO e.g. the posterior samples of the parameters, not a knowledge graph.

usecases/MinimumWorkingExample/dodo.py
```diff
@@ -18,9 +18,6 @@
```
joergfunger, line 18: The Youngs modulus should be computed for all analyzed files in this setup and the results stored in the KG.

usecases/SimpleOntology/Mapping/SimpleOntologyMappedProbe1.ttl
```diff
@@ -32,9 +32,13 @@ xsd:float a rdfs:Datatype .
 
 xsd:string a rdfs:Datatype .
```
joergfunger, line 33: not sure if the changes in these files here are on purpose, I guess they are not.

joergfunger (changes requested):
I had the impression that many of the comments where just marked as resolved without actually resolving it. When a comment is not to be resolved (e.g. when making another issue for that), this should be marked. I will schedule a meeting to discuss that.

lebedigital/calibration/forwardmodel_linear_elastic_cylinder.py
```diff
@@ -43,28 +40,33 @@ def response(self, inp: dict) -> dict:
         # input parameters
         parameters['E'] = inp["E"]
         parameters['nu'] = inp["nu"]
-        parameters['height'] = inp["height"]
         parameters['radius'] = inp["radius"]
+        parameters['height'] = 100  # gauge length in experiment in mm
```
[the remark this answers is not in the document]
eriktamsen: I checked the extracted metadata and it is not written there..., not sure where this data would be stored currently.

lebedigital/calibration/forwardmodel_linear_elastic_cylinder.py
eriktamsen: This approach currently only works as we deal with a linear material model, so storing this would not be a general solution/method.

lebedigital/calibration/utils.py
```diff
@@ -0,0 +1,238 @@
...
+        "exp_name": exp_name,
+        "height": float(results["length"]),
+        "diameter": float(results["diameter"]),
+        "displacement": np.array(df["displacement"]),
+        "force": np.array(df["Force [kN]"]),
+    }
+    return output
+
+
+class PosteriorPredictive:
```
eriktamsen, line 110: I would not expect this class definition to be in "utils.py", I would suggest to have clear/more thought through structure for the script that will allow for extension to the demonstrator

tests/emodulus_calibration/test_emodulus_calibration.py
```diff
@@ -0,0 +1,31 @@
...
+                                       calibration_metadata={"E_loc": E_loc, "E_scale": E_scale},
+                                       calibrated_data_path=data_path, mode='cheap')
+
+    # checking if the KG file for the calibration is created
+    assert (data_path / f'calibrationWorkflow{input_file}.owl').is_file()
+
+    # checking for the mean of the calibrated E modulus
+    # TODO (Atul), there should be a way to make sure the result is the same each time? Defining a random seed?
+    #      rel = 0.4 is too large...
+    assert np.mean(E_samples) == pytest.approx(E_loc, rel=0.4)
```
eriktamsen, line 29: is there a way to define maybe a random seed or something, such that the result of the estimation will always give the same answer, even with a small number of samples?

usecases/SimpleOntology/Mapping/SimpleOntologyMappedProbe1.ttl
eriktamsen: I will have a look, might have been me, when I merged from main together with Atul

eriktamsen (commented):
I just added some of the points we already talked about but which are not yet implemented.

usecases/MinimumWorkingExample/dodo.py
eriktamsen: I would suggest to use the "cheap" or "single file" option here as default, so that the CI tests are not taking forever, but implement a "full" mode to locally check that there are not problems with specific files.

usecases/MinimumWorkingExample/dodo.py
eriktamsen: I copied the version of the dodo file from main branch. this was also part of the merge problem.

.github/workflows/lebedigital.yml
```diff
@@ -48,7 +48,7 @@ jobs:
         shell: bash -l {0}
         run: |
           cd $GITHUB_WORKSPACE/tests/
-          pytest -s -W error::UserWarning --ignore=ModelCalibration/tests/simulation -m "not login"
+          pytest -s -W error::UserWarning
```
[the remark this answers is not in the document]
eriktamsen: fixed it

lebedigital/calibration/utils.py
```diff
@@ -0,0 +1,238 @@
...
+        # Monte carlo step
+        output = []
+        for i in range(0, samples):
+            # TODO improve this!!!, this is specific to "three point bending beam"
+            parameters = fenics_concrete.Parameters()  # using the current default values
+            # input values for the material
+            parameters['E'] = self._parameter[i]
+            parameters['nu'] = self._known_input
+            y = self._forward_solver(parameters)
+            output.append(y.magnitude)  # removes the units
```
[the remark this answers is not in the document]
eriktamsen: I dont want to remove them, this was the quick and dirty way to actually get the script to run with out having to change things I dont fully understand.
Carrying them though the whole process in a consistent way requires changes to the setup that need to be performed by @atulag0711, or at least as a joint effort. 
I suggest to move the correct implementation of units to a new issue, as this is also related to how the data is passed and stored.

tests/emodulus_calibration/test_emodulus_calibration.py
eriktamsen: I removed them

tests/emodulus_calibration/test_emodulus_calibration.py
eriktamsen: I removed them

usecases/MinimumWorkingExample/dodo.py
eriktamsen: As a note, this is not actually outdated!!! The file has changed (for other reasons) but this point is still valid.

usecases/SimpleOntology/Mapping/SimpleOntologyMappedProbe1.ttl
eriktamsen: Fixed it

tests/emodulus_calibration/test_emodulus_calibration.py
atulag0711: Yes I will take a look.

lebedigital/calibration/utils.py
atulag0711: Yes good point. In demonstrator this can be/will be used as follows: After we get x^* (the optimized design variable), we would get the prediction of the KPIs +- sd. Essentially running your snakemake several times, for different sample of (hydration parameters and paste properties E and strength).

tests/emodulus_calibration/test_emodulus_calibration.py
atulag0711: This is outdated after our discussion.

lebedigital/calibration/posterior_predictive.py
```diff
@@ -0,0 +1,66 @@
...
+    # is harcoded here.
+    # three_point = three_point_bending_example()
+    pos_pred = PosteriorPredictive(
+        forward_solver, known_input_solver=nu, parameter=E_pos
+    )
+
+    if mode == "cheap":
+        mean, sd = pos_pred.get_stats(samples=3)  # mean : ~365 N/mm2, sd = 30
+    else:
+        mean, sd = pos_pred.get_stats(samples=10)  # mean : ~365 N/mm2, sd = 30
```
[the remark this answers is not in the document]
atulag0711: Yes valid point, I will add a parameter dict which allows someone to change from outside.

lebedigital/calibration/posterior_predictive.py
```diff
@@ -0,0 +1,66 @@
...
+
+# local imports (probeye)
+from probeye.ontology.knowledge_graph_import import import_parameter_samples
+
+# local imports (others)
+from lebedigital.simulation.three_point_bending_beam import three_point_bending_example
+from lebedigital.calibration.utils import PosteriorPredictive
+from lebedigital.unit_registry import ureg
+
+def perform_prediction_three_point_bending(knowledge_graph_file:str,forward_solver=three_point_bending_example,mode = 'cheap'):
```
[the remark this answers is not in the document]
atulag0711: This is outdated after our discussion.

lebedigital/calibration/utils.py
```diff
@@ -199,5 +180,58 @@ def get_stats(self, samples: int) -> tuple:
         return mean, sd
 
     def plot_posterior_predictive(self):
-        raise NotImplementedError(
-            "...")
+        raise NotImplementedError("...")
+
+
+def extract_third_load_cycle(
```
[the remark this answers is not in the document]
atulag0711: This is done already.

eriktamsen:
@atulag0711, please don't forget to finish this PR. I would like to add things (VMAP output) to the predictive simulation, or maybe to a single final run. To do that I need this branch in main.

atulag0711:
> @atulag0711, please don't forget to finish this PR. I would like to add things (VMAP output) to the predictive simulation, or maybe to a single final run. To do that I need this branch in main.

Yes I will work on it tomorrow.

tests/emodulus_calibration/test_emodulus_calibration.py
atulag0711: Updated. The seed in emcee sampler somehow is not changing the sampling behaviour. but yes the rel is reduced.

atulag0711:
> @atulag0711, please don't forget to finish this PR. I would like to add things (VMAP output) to the predictive simulation, or maybe to a single final run. To do that I need this branch in main.

Alright done. Updated the files. Nowhere I am querrying any KG. Also posterior predictive is general again. Just needed to write a wrapper for your three point bending test.

Please merge. If further code abstraction is desired if any case, please feel free to change as per your needs.

atulag0711:
Somehow the dodo file from the main is not working here. Dont know why. All the pytests do work though.

eriktamsen:
outdated branch, changes are now worked on in #152

PR #60, opened 2022-10-04
59 speed up building conda environments in GitHub actions

joergfunger:
Speed up the building of conda environments by 
1. using mamba (as already done)
2. caching versions once per day (they are only rebuild if either the environment file or the date changes)
3. joining the different task into subsequent steps (to avoid building the same environment twice)
The total time when using a cached version is now reduced to 2:42 (from approximately 10min), if a cached conda environment can be found. A new version can be forced by increasing the CACHE_NUMBER. The key for recreating an environment is actually a hash from the environmentfile, the CACHE_NUMBER and the DATE.
In addition, a cronjob is added to run the actions every day at 3:15am, such that a prebuild version of the environment should usually be available when pushing a new version.

### DIFF
--- a/.github/workflows/lebedigital.yml
+++ b/.github/workflows/lebedigital.yml
@@ -14,7 +14,7 @@ on:
     - cron: '3 15 * * *'
 
 env:
-  CACHE_NUMBER: 0  # increase to reset cache manually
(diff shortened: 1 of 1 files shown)


eriktamsen (approved):
Looks good, thanks for the changes. This should be merged.

PR #62, opened 2022-10-10
61 creating initial optimization problem

eriktamsen:
The main point was to add a column simulation as a test case for @atulag0711's optimization.
In addition I added a script for the planned homogenization method and updated the graph structure.
Tests have been written both for the column and the homogenization script.


tests/demonstrator_scripts/test_column_simulation.py
```diff
@@ -0,0 +1,58 @@
...
+    # p['Q_inf'] = self.Q_pot * self.density_binder * self.b_ratio  # potential heat per concrete volume in J/m3
+    parameters['B1'] = 2.916E-4  # in 1/s
+    parameters['B2'] = 0.0024229  # -
+    parameters['eta'] = 5.554  # something about diffusion
+    parameters['alpha_max'] = 0.875  # also possible to approximate based on equation with w/c
+    parameters['E_act'] = 5653 * 8.3145  # activation energy in Jmol^-1
+    parameters['T_ref'] = 25  # reference temperature in degree celsius
+
+    # simulation time
+    full_time = 1  # simulation time in hours
```
joergfunger, line 50: Just for ease of use, I guess it might be easier not to change the units of time.  I would rather have everything stored in seconds, and if someone is really interested in h, then it might be easier to convert it just before printing (time/3600).

lebedigital/simulation/precast_column.py
```diff
@@ -0,0 +1,43 @@
...
+
+        if pv_output:
+            problem.pv_plot(t=t)
+
+        # prepare next timestep
+        t += dt
+
+    print('Done!')
+    # return lists with time steps, max temperature, max yield
+    return problem.sensors['MaxYieldSensor'].time, problem.sensors['MaxTemperatureSensor'].data, problem.sensors['MaxYieldSensor'].data
```
joergfunger, line 43: could we return a pandas df with time, and then the data from each sensor as another column?

joergfunger (approved):
Just minor changes, thanks a lot. Can be merged without further review from my side.

lebedigital/simulation/concrete_homogenization.py
```diff
@@ -0,0 +1,66 @@
...
+
+    this function calls the Mori-Tanaka homogenization scheme
+    A dictionary with the required parameter is given
+
+    Parameters
+    ----------
+    parameters : dict
+        List with all required parameters
+        Paste properties
+        - paste_E
```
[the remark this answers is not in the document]
eriktamsen: Done

tests/demonstrator_scripts/test_column_simulation.py
eriktamsen: I agree with you that it is usually  better/cleaner to use the same units everywhere.
However, here I chose hours and minutes as input as this would all a intuitive setting of the values (this is by the way the origin of the discussion on using seconds as general unit everywhere).
I don't think there is a right way, so I changed it now back to seconds.

lebedigital/simulation/precast_column.py
eriktamsen: done

PR #63, opened 2022-10-12
Add files via upload

JulietteWinkler:
Here is the ontologie for the minimum working example.


joergfunger (approved):
Thanks for the ontologies. Could you add another Readme.md file in the ontology folder that explains what each ontology is for. (MiWoEx is not self explaining). In addition, are the other ontologies in that folder relevant or can we remove those. If yes, could you delete them?

joergfunger:
Could you please add a Readme.md. In addition, is the name MiWoEx the final name? We would like to use the same ontology for the demonstrator, so we should at least have a more general name that covers that. Why not e.g. "concrete".

joergfunger:
@JulietteWinkler  I tried to merge your changes into main. There are a couple of problems. One is related to the fact that in the mapping procedure some of the deleted ontologies are still imported e.g. MergedAllCoreOntology.owl , EM.xml, ..(see [here](https://github.com/BAMresearch/LebeDigital/blob/19405e933a5b26d5aafb8237e12f9c6701338960/lebedigital/mapping/emodul_mapping.py#L116)) There is also the mseo_mid that is loaded, but owlready requires this in owl format (not in ttl as given here). Could you please discuss with @AidaZt or (@soudehMasoudian  @alFrie) to fix this. Currently, the tests are not working and thus I can't merge into the main branch.

PR #67, opened 2022-10-14
64 correction of raw mixture data excel files

alFrie:
Three raw data files for the mixture have been corrected (labels were wrong) and one file has been deleted since its layout and labels contained mistakes.


joergfunger (approved):
Thanks.

PR #68, opened 2022-10-14
ravi patel

raviapatel:
Added script for beam design


lebedigital/BeamDesign/BeamDesign.py
```diff
@@ -0,0 +1,166 @@
...
+
+    Parameters
+    ----------
+    span : float
+        beam span.
+
+    Returns
+    -------
+    tuple
+        tuple of breath and depth of the beam.
```
joergfunger, line 21: what is breath? height?

lebedigital/BeamDesign/BeamDesign.py
```diff
@@ -0,0 +1,166 @@
...
+    """
+    function to compute max bending moment and shear force for simply supported beam with point load or distributed load
+
+    Parameters
+    ----------
+    span : float
+        beam span.
+    load : float
+        beam load.
+    type : str
```
joergfunger, line 40: why don't you specify both a point load (add documentation that this is in the center) and a distributed load. If one is intereste in one or the other, they could just set the corresponding values. What are the units?

lebedigital/BeamDesign/BeamDesign.py
```diff
@@ -0,0 +1,166 @@
...
+        whether point_load or distributed_load to describe type of load.
+
+    Returns
+    -------
+    tuple
+        tuple of  maximum moment and maximum shear force.
```
joergfunger, line 46: units?

lebedigital/BeamDesign/BeamDesign.py
```diff
@@ -0,0 +1,166 @@
...
+                       cover:float,
+                        ) -> dict:
+    """
+    Function to design singly reinforced beam with minimum shear reinforcement required.
+
+    Parameters
+    ----------
+    span : float
+        Span of the beam in mm.
+    MaxMoment : float
```
joergfunger, line 76: b and d are parameters but are missing in the documentation.

lebedigital/BeamDesign/BeamDesign.py
```diff
@@ -0,0 +1,166 @@
...
+    out = {}
+    out["Depth[mm]"]=d
+    out["cover[mm]"] = cover
+    out["width[mm]"] = b
+    out["bottom_steel_dia[mm]"] = steelDia
+    out["bottom_steel_numbers"] = nsteel
+    out["top_steel_dia[mm]"] = steelDia
+    out["top_steel_numbers"] = 2
+    out["shear_reinforcement_dia[mm]"] = 12
+    out["shear_reinforcement_spacing[mm]"] = S
```
joergfunger, line 164: Not sure what the output really is. It would be good if that is not a discontinous value (in terms of number of steel bars), but rather the required surface (which could then be compared with the given one). This way, the optimization can compute meaningful gradients.

tests/beam_design/test_beam_design.py
```diff
@@ -0,0 +1,24 @@
...
+    out=  beam_section_design(span =6500,
+                              b = b,
+                              d = d,
+                              MaxMoment=MaxMoment,
+                              MaxShearForce=MaxShearForce,
+                              fck=20,
+                              fyk=500,
+                              steelDia=12,
+                              cover=25)
+    for key,val in out.items():
```
joergfunger, line 19: this is not a test, you would have to check the values, not only print that.

joergfunger (changes requested):
Thanks for the script, very good starting point. Couple of improvements would be nice before merging.

tests/beam_design/test_beam_design.py
```diff
@@ -0,0 +1,24 @@
+
+from lebedigital.BeamDesign import BeamDesign
+from BeamDesign import *
```
eriktamsen: import * is not encouraged, see e.g.: [why import star in python is a bad idea](https://www.geeksforgeeks.org/why-import-star-in-python-is-a-bad-idea/)
better import BeamDesign and then call the functions as BeamDesign.functionName, or specify specifically which functions you need.

lebedigital/BeamDesign/BeamDesign.py
```diff
@@ -0,0 +1,166 @@
...
+    -------
+    tuple
+        tuple of breath and depth of the beam.
+
+    """
+    #span/depth ratio for simply supported beam is 15 Therefore,
+    d = 50*round((span/15)/50)
+    #width of beam
+    b = 0.5*d
+    return b,d
```
eriktamsen, line 28: Maybe we need a general discussion on this but I don't think this is what we are looking for.
I understand that the EC might recommend a ratio of span to depth (width?) of 15, but are you required to do so?
I would prefer to directly optimize height and width, otherwise our plan of a geometry optimization is pointless.

lebedigital/BeamDesign/BeamDesign.py
```diff
@@ -0,0 +1,166 @@
...
+        tuple of breath and depth of the beam.
+
+    """
+    #span/depth ratio for simply supported beam is 15 Therefore,
+    d = 50*round((span/15)/50)
+    #width of beam
+    b = 0.5*d
+    return b,d 
+
+def max_bending_moment_and_shear_force(span:float,load:float,load_type:str) ->tuple:
```
eriktamsen, line 30: I would prefer to directly use width and height as input here. (See previous comment)
In addition, if I understood correctly we are also thinking about including options of I and T beams right?
Would it be possible to parametrize the beam cross section to include those options already (the rectangular one just being a special case of this)?

eriktamsen (changes requested):
Thanks for the code. 
I am not sure if the current computation of the bending moment and shear force,is sufficient for our goals.

lebedigital/BeamDesign/BeamDesign.py
raviapatel: i will make it consistent width and depth

lebedigital/BeamDesign/BeamDesign.py
raviapatel: i thought this is much a easier and one can then do superposition  on their own in the script which one has to execute for optimization not that this is just functions to support execution and execution script has to be built using those function. This is how i understood one codes. If not then we have to define a problem and then we can make specific execution script

lebedigital/BeamDesign/BeamDesign.py
raviapatel: i will add them

lebedigital/BeamDesign/BeamDesign.py
raviapatel: i made this as output for entire design of beam. I am not sure if gradient based optimization is best for this kind of systems because there is always dimensional rounding and you will have lot of local minimas. You should look into that. I have this as an output because it gives me complete design of section while defining output in execution script one can delete or ignore certain variables. Again I emphasize there are helper functions and not the script which you will attach in workflow for script you need to provide problem definition

tests/beam_design/test_beam_design.py
raviapatel: you are right but check with what??? for now i can make such that it checks with the output it is currently computing

tests/beam_design/test_beam_design.py
raviapatel: i will change that

lebedigital/BeamDesign/BeamDesign.py
raviapatel: this function you can use for initial guess and then you can optimize further with height or width

lebedigital/BeamDesign/BeamDesign.py
raviapatel: you can directly use that i think there is misunderstanding this is not execution script this are helper functions... 
The design of beam is only valid for singly reinforced rect beam not for I and T beam so those sections wont work. In eurocode 2 if you check for singly reinforced T beam they are designed as rect beam with contribution from flange not computed. Hence they will be always more expensive. For I beam i couldnt find any design guidelines but must then ask someone. But if the section is not designed for compression reinforcement which is the case for singly reinforced section then I beam would also not make any different in terms of amount of steel. As far as I understood we dont want to go towards compression reinforcement or heavily reinforced section for design in final demostrator..

lebedigital/BeamDesign/BeamDesign.py
eriktamsen: I think including the superposition within this function is much more user friendly.

lebedigital/BeamDesign/BeamDesign.py
eriktamsen: Now I understand. So this can be used as an initial guess.

lebedigital/BeamDesign/BeamDesign.py
eriktamsen: It would be helpful if you then could add the execution script, so that we can understand how these will work together.

eriktamsen: I am not sure in what way the T beam is "allways more expensive". Is it not the point to reduce the amount of used cement? I would assume this could be a good way.
I would assume that for a simple bending load the I beam will be computed the same as a T beam. A difference will be in rotational moments, but I dont think we can about that load case.

raviapatel: you are right stress distribution would be different for T beam within section but way eurocode suggest to design is that dont consider flange and thus you reduce T beam to rectangular beam. Hence for design purposes the extra concrete in flange has no role and hence you will get same reinforcement as one for rectangular beam. There can be more detail design with compression reinforcement considered in the section but that is not done in the present script. It is only for special case of singly reinforced concrete beam with rectangular c/s

eriktamsen: I looked at the test script, this makes it more clear. It would be better to write a function, basically doing what the test is doing and then testing this function. That way we can have a clear overview of the input and output of the final function and directly use it in the workflow.
After a quick search I found there are different cases, according to e.g. (https://rudolfbaumgart.de/scripte/biegebem.pdf). I guess this discussion might be easier in person. For now the rectangle is a good start.

raviapatel: > 

Thanks for the ref i see they divided into two cases case 1 they dont consider flange and case 2 where they change to width of the c/s to an equivalent width and then proceed with rectengular section design for now i would suggest to make first series of optimization on rect section and see if something logical is coming out.

eriktamsen: @raviapatel any progress?

joergfunger (changes requested):
Thanks for the new updates. Looks almost finished, just a couple of additional small changes.

lebedigital/BeamDesign/BeamDesign.py
```diff
@@ -27,55 +27,53 @@ def section_dimension_rule_of_thumb(span:float)->tuple:
...
                         ) -> dict:
     """
     Function to design singly reinforced beam with minimum shear reinforcement required.
 
     Parameters
     ----------
-    span : float
-        Span of the beam in mm.
-    b: float
+    span : int
```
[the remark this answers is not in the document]
eriktamsen: fixed

lebedigital/BeamDesign/BeamDesign.py
```diff
@@ -166,5 +164,58 @@ def beam_section_design(
...
+    -------
+    float
+        difference of  specified area and required area given the diameter of steel and number of steel bars in the bottom of the section.
+        It is negative if  design is not satisfied and positive if design is satisfied. Optimal will be close to zero.
+    """
+    MaxMoment,MaxShearForce=max_bending_moment_and_shear_force(span,point_load,distributed_load)
+    design = beam_section_design(span,b,d,MaxMoment,MaxShearForce,fck,fyk,steelDia,cover)
+    specified_area = (math.pi*steelDia**2/4) * n_bottom
+    required_area = design["area_bottom_steel[mm^2]"]
+    return specified_area-required_area
```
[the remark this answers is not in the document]
eriktamsen: done

lebedigital/BeamDesign/BeamDesign.py
```diff
@@ -166,5 +164,58 @@ def beam_section_design(
     out["top_steel_numbers"] = 2
     out["shear_reinforcement_dia[mm]"] = 12
     out["shear_reinforcement_spacing[mm]"] = S
+    out["area_bottom_steel[mm^2]"] = As1
```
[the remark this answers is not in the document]
eriktamsen: done

eriktamsen (approved):
looks fine to me now

tests/beam_design/test_beam_design.py
```diff
@@ -9,20 +9,31 @@ def test_beam_design():
...
                               MaxShearForce=MaxShearForce,
                               fck=20,
                               fyk=500,
-                              steelDia=16,
+                              steelDia=12,
                               cover=25)
     for key,val in out.items():
         print(f"{key} --> {val}")
-    assert(out["bottom_steel_numbers"]==2)
+    assert(out["bottom_steel_numbers"]==3)
```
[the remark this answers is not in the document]
joergfunger: Maybe I'm looking in the wrong file, but in test beam design only bottom_steel_numbers is tested, the ratio of required area of steel to given is not tested, which I guess we might use for the optimization (since I thought we wanted to prescribe the number of steel bars and then only having the other concrete parameters to be flexible in the design.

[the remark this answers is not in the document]
eriktamsen: You are looking at an outdated file.
[This](https://github.com/BAMresearch/LebeDigital/blob/66-beam-design/tests/beam_design/test_beam_design.py) is the current version, which only tests the difference of area as the final result, due to the reasons you have explained ;)

joergfunger (approved):

PR #70, opened 2022-10-17
fixed parameter name

eriktamsen:
name changed from "boundary" to "T_bc1"

### DIFF
--- a/tests/demonstrator_scripts/test_column_simulation.py
+++ b/tests/demonstrator_scripts/test_column_simulation.py
@@ -29,7 +29,7 @@ def test_column_simulation():
 
     # temperature setings:
     parameters['T_0'] = 40  # initial temperature of concrete
-    parameters['T_boundary'] = 40  # constant boundary temperature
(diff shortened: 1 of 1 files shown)


tests/demonstrator_scripts/test_column_simulation.py
```diff
@@ -29,7 +29,7 @@ def test_column_simulation():
 
     # temperature setings:
     parameters['T_0'] = 40  # initial temperature of concrete
-    parameters['T_boundary'] = 40  # constant boundary temperature
+    parameters['T_bc1'] = 20  # constant boundary temperature
```
joergfunger, line 32: what is bc1 (I guess there is another bc2?

eriktamsen: this goes back to more complex test cases with different temperatures at different positions. this is however coded in fenics_concrete, so not something we would change here.

atulag0711 (approved):

PR #73, opened 2022-10-27
Changes at the mixture metadata extraction script

alFrie:
I changed the script in a way that any missing label (still using a predefined selection of necessary labels and comparing the found labels to this) will raise an error. Loguru will catch this error, which can be then also be seen in a created logging-file and just pass the script without breaking the workflow. No outputfile will be created in this case.

Edit: I also fixed a mistake where more than 2 additions would just overwrite previous additions. Now the script will break if more than two additions are found.


eriktamsen (changes requested):
Thank you so much. This looks great.
I am sorry but we need one more small step before we can merge.
Please add the extraction function to the dodo file in usecases/MinimumWorkingExample
I lost a bit the overview, the unit tests this code are being developed in a separate branch? If not, it would be important to add a test script to the test directory as well.

eriktamsen:
Yes, I guess that would be the easiest way.
It would be good if the chosen example mix is the mix that is used in 'Wolf 8.2 Probe 1' (it is 2014_12_10 Wolf.xls). If for some reason we dont have data on that mix or that dataset has problems, you can also define a different sample for the easy example.
The data is stored in [mix.dat](https://github.com/BAMresearch/LebeDigital/blob/main/usecases/MinimumWorkingExample/Data/E-modul/Wolf%208.2%20Probe%201/mix.dat)
If you can think of an easy way to automatically choose the corresponding mix, for the chosen test data, that would be really nice, but I guess once everything is working, we can also improve our doit file.

lebedigital/raw_data_processing/mixture/mixture_metadata_extraction.py
```diff
@@ -140,6 +142,7 @@ def extract_metadata_mixture(
             if missing_labels == ['Zusatzstoff2']:
                 logger.warning('No addition2 in raw data.')
             else:
+                logger.error('Check raw data, there are labels missing.')
```
joergfunger, line 145: The error is not fully correct. If you have two errors (one resulting from a real missing label, and one from Zusatzstoff2 missing), you would only return the debug message instead of an error.

joergfunger: sorry, my mistake

joergfunger:
> 

Not sure if I understand correctly, are there multiple mixtures in a single excel sheet?

joergfunger:
This mix seems not to appear in the mixes used the tests (compression, E). As such, I would suggest excluding that file directly when building the list of files in the dodo file. (all files minus that specific one). If we realize at same point that there is a mix missing in the tests, we could think about how to handle that later.

joergfunger:
As for the mix missing in some cases, there are tests that are based on real structures with drilled cores where the mixture is not available. This should not result in an error, but rather in an empty mix. Not sure though how we handle that in the queries, but that is also to be discussed later.

joergfunger:
As for the mix file with multipe sheets, exclude it in the dodo file. Also, throw an error if there are multiple sheets. The naming should be kept as the my.xls -> my.yaml.

joergfunger:
As for the relation between E and mix, there should be a mix.dat in each test file that includes the name of the mix sheet. Could you add that name to the metadata.yaml of the E metadata?

usecases/MinimumWorkingExample/dodo.py
```diff
@@ -23,18 +26,27 @@
...
 raw_data_emodulus_directory = Path(ParentDir, 'Data', 'E-modul')  # folder with folders of raw data files
 metadata_emodulus_directory = Path(emodul_output_directory, 'metadata_yaml_files')  # folder with metadata yaml files
 processed_data_emodulus_directory = Path(emodul_output_directory, 'processed_data')  # folder with csv data files
 knowledge_graphs_directory = Path(emodul_output_directory, 'knowledge_graphs')  # folder with KG ttl files
 
+# defining paths for mixture
+raw_data_mixture_directory = Path(ParentDir, 'Data', 'Mischungen')  # folder with raw data files (excel)
+mixture_output_directory = Path(ParentDir, 'mixture')  # folder with folders
+metadata_mixture_directory = Path(mixture_output_directory, 'metadata_yaml_files')  # folder with mixture metadata yaml files
+#processed_data_mixture_directory = Path(mixture_output_directory, 'processed_data')  # folder with csv data files
```
joergfunger, line 40: delete that if not needed

usecases/MinimumWorkingExample/dodo.py
```diff
@@ -23,18 +26,27 @@
...
+mixture_knowledge_graphs_directory = Path(mixture_output_directory, 'knowledge_graphs')  # folder with KG ttl files
+
 # when "cheap option" is run, only this souce of raw data is processed
 cheap_example_name = 'Wolf 8.2 Probe 1'
```
joergfunger, line 44: For the mix, I would not introduce a cheap example that is very fast and all mixes can always be run in the tests.

usecases/MinimumWorkingExample/dodo.py
```diff
@@ -23,18 +26,27 @@
...
+cheap_mixexample_name = '2014_12_10 Wolf.xls' # corresponding mixture file
 
 # create folder, if it is not there
 Path(emodul_output_directory).mkdir(parents=True, exist_ok=True)
+Path(mixture_output_directory).mkdir(parents=True, exist_ok=True)
```
joergfunger, line 49: could you group the stuff related to mix and E (such that all emodul dirs are in one block, and then the mix is another one)

usecases/MinimumWorkingExample/dodo.py
```diff
@@ -87,6 +99,31 @@ def task_extract_processed_data_emodul():
             }
 
 
+#extract standardized metadata for the mixture
+def task_extract_metadata_mixture():
+    # create folder, if it is not there
+    Path(metadata_mixture_directory).mkdir(parents=True, exist_ok=True)
+
+    # setting for fast test, defining the list
+    if config['mode'] == 'cheap':
```
joergfunger, line 108: see above, IMO no need to have a cheap example here (for the E this is primarily for the parameter estimation, the extraction is also there quite straightforward)

joergfunger (approved):
Thanks, that looks good and there is only one final wish from my side. The dodo file gets complex with all the data being imported. As such, I guess we will have to introduce nested dodo files at some point. As soon as we introduce the link between E and mix, the mix has to be processed first.

joergfunger:
Could you rebase on the current main branch, the tests have been fixed and now your tests should run as well in the CI.

usecases/MinimumWorkingExample/dodo.py
joergfunger: at some point we would have to separate that also in different task (with the mix being a precondition for the Youngs modulus), so it would be great to do that already now.

joergfunger:
The problem seems to be in the metadata extraction part. Did it work for you when running pytest in the test directory? Did you use the conda environment provided on the main level?

joergfunger:
Just checked it myself, xlrd  and openpyxl are missing in the environment.yaml file on the main level.

joergfunger:
No, I did not add anything.

eriktamsen (approved):
Looks good to me. Thank you!
I added the option to run "doit mode=single" to only do one file. not so much to save computation time (as in "doit mode=cheap") but to have a better overview of the outputs, this is meant as a development/debugging mode.

PR #77, opened 2022-11-10
fix test cylinder simulation and rm aux files

joergfunger:
changed the clinder simulation to free bc and compare with the exact solution with an additional scaling factor to compensate for the dscretization error due to the very low number of elements.


eriktamsen (commented):
I am not sure why the fest started to fail, but the changes look fine.

eriktamsen (approved):
I am not sure why the fest started to fail, but the changes look fine.

PR #82, opened 2022-11-22
improved parameter name and location of test

eriktamsen:
changed location of test file
changed parameter names for better usability


joergfunger (approved):
The question is, who is going to read that. For someone not familiar with the Eurocode, those suggestions improve the readibility, for someone familiar with that fck is the characteristic (k) strength of concrete in compression (fct,k) would e.g. be the one in tension. And y is for steel (don't ask me why). Since we are hoping to allow anyone to read these files, the suggestions help.

PR #85, opened 2022-11-25
83 initialization of opimization paper

eriktamsen:
This is my suggestion for the initial structure of the paper directory. 
Don't judge the content yet, it was only copied as example.
I would suggest to merge this with main as a start.
Everyone can create their own paper branch and work there and do a PR from time to time to get feedback from the others.
We can of course also merge from one individual paper branch to the other if we need to do a quick update without going through a PR.
I included a short readme in the directory to give a quick overview of the automation features currently implemented.
Looking forward to the progress.


usecases/optimization_paper/README.md
```diff
@@ -0,0 +1,38 @@
...
+The current proposal is the write the paper in the **tex** folder and split the document in sections, which are located in **tex/tex_sections**.
+Modules like **workflow_graph** can be added as directory to the main directory and the included in the pydoit workflow.
+Figures generated for the paper are located in **figures**.
+
+## Features
+
+### pydoit
+There is a **dodo.py** file which includes all steps required to generate the paper.
+Expensive computations should not be always. We will need to find a good solution.
+I would propose to use a similar setup as in the MinimumWorkingExample and us a call like
```
joergfunger, line 15: and us call (check wording)

usecases/optimization_paper/README.md
```diff
@@ -0,0 +1,38 @@
...
+There is a **dodo.py** file which includes all steps required to generate the paper.
+Expensive computations should not be always. We will need to find a good solution.
+I would propose to use a similar setup as in the MinimumWorkingExample and us a call like 
+```
+doit mode=full
+```
+to trigger the full calculation. But I am open to suggestions. Maybe it is not needed.
+
+### Github action
+The pydoit workflow is included in the actions so we have a direct feedback is everything is working as planned.
```
joergfunger, line 22: check wording (is everything is working)

usecases/optimization_paper/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+### tex-macros
+In **tex/macros/tex_macros.tex** are macros defined which are static.
+This whould be for example the definition of varibales in equations, for example
```
joergfunger, line 26: varibales -> variables

usecases/optimization_paper/README.md
```diff
@@ -0,0 +1,38 @@
...
+```
+\newcommand{\DOH}{\alpha} % degree of hydration
+```
+
+### py-macros
+There is the possiblity to generate tex-macros by a python script.
+The allows variable input to be included in the paper.
```
joergfunger, line 33: The -> This

usecases/optimization_paper/README.md
```diff
@@ -0,0 +1,38 @@
...
+Figures generated for the paper are located in **figures**.
+
+## Features
+
+### pydoit
+There is a **dodo.py** file which includes all steps required to generate the paper.
+Expensive computations should not be always. We will need to find a good solution.
+I would propose to use a similar setup as in the MinimumWorkingExample and us a call like 
+```
+doit mode=full
```
joergfunger, line 17: I would rather propose it the other way round, if we are interested in a subset to exclude something, I would add a ci option

usecases/optimization_paper/dodo.py
```diff
@@ -0,0 +1,65 @@
+import pathlib
+from workflow_graph.paper_workflow_graph import paper_workflow_graph
```
joergfunger, line 2: What is that? (I stumbled upon this in the first task). Is that the diagramm of the workflow automatically being generated?

usecases/optimization_paper/dodo.py
```diff
@@ -0,0 +1,65 @@
...
+        "actions": [(py_macros, [py_macros_file.with_suffix('')])],
+        "targets": [py_macros_file],
+        "clean": True,
+    }
+
+
+def task_paper():
+    """compile pdf from latex source"""
+    return {
+        "file_dep": [paper_file, workflow_output_file, py_macros_file],
```
joergfunger, line 61: The list of dependencies might get large here. One option we tried for a previous paper was that the paper_inputs (dependencies) were collected in all tasks (so each task that produced an input to the paper was itself responsible for adding that to that dict) and then the task paper was recreated after all other tasks are executed using this previously created dict. This has the advantage that you don't have to touch the dependencies here everytime a new figure is added. It is only a suggestion, feel free to ignore.

usecases/optimization_paper/tex/optimization_paper.tex
```diff
@@ -0,0 +1,122 @@
...
+\author[2]{\fnm{Jörg F.} \sur{Unger}}\email{joerg.unger@bam.de}
+
+\affil*[1]{\orgdiv{Data-driven Materials Modeling}, \orgname{Technische Universität München}, \orgaddress{\street{Boltzmannstraße 15}, \city{Garching}, \postcode{85748}, \country{Germany}}}
+
+\affil[2]{\orgdiv{Modeling and Simulation}, \orgname{Bundesanstalt für Materialforschung und -prüfung}, \orgaddress{\street{Unter den Eichen 87}, \city{Berlin}, \postcode{12205}, \country{Germany}}}
+
+
+\input{tex_sections/abstract.tex}
+
+\section{Introduction}\label{sec:introduction}
```
joergfunger, line 86: I did this structuring of a thesis into sections as well, not sure if that is needed for a paper, since with the structure being displayed in most editors it might be as simple as navigating to an additional file. But I guess for the initial commit with independent contributions this somehow decouples everyones work (so just a comment, potentially also to be ignored).

usecases/optimization_paper/tex/optimization_paper_bibliography.bib
```diff
@@ -0,0 +1,147 @@
+%% Journal article
+@article{bib1,
```
joergfunger, line 2: I usually add the name of the first author and year (so \cite{Tamsen2023}, potentially adding the abbreviated journal name. This makes reading that in the source code a little easier and also allows to reuse one bib in another one.

usecases/optimization_paper/tex/optimization_paper.tex
```diff
@@ -0,0 +1,122 @@
...
+\input{tex_sections/introduction.tex}
+
```
joergfunger, line 88: Should we distinguish between the state-of-the-art and an introduction/overview/goals of the paper. I guess we could put both in the introduction, but I don't think the second part will just be 5 lines. It might also be a good idea that we start with the objective part, because that would allow all the other parts to concentrate on these objectives.

joergfunger (approved):
Thanks for making the initial step. Just some minor comments, feel free to merge or ignore any of the remarks.

usecases/optimization_paper/dodo.py
eriktamsen: YEs

usecases/optimization_paper/dodo.py
eriktamsen: @joergfunger: Is the code of the mentioned paper public, or at least accessible to me?

usecases/optimization_paper/README.md
eriktamsen: I will setup the possibility to implement different options, we can then always decide how to use them.

usecases/optimization_paper/tex/optimization_paper.tex
eriktamsen: I think it will be useful when working on it via Git, as we otherwise would usually work in the same file which can easily lead to annoying problems when merging.

usecases/optimization_paper/tex/optimization_paper_bibliography.bib
eriktamsen: This is still just part of the template, will be improved.

usecases/optimization_paper/tex/optimization_paper.tex
eriktamsen: This is not supposed to be the suggestion for the actual structure. We should have a discussion together about that.

usecases/optimization_paper/dodo.py
eriktamsen: Done

PR #91, opened 2022-11-30
87 mi wo ex

soudehMasoudian:
added required files into the correct package (ConcreteOntology)


eriktamsen:
Why are there so many files here (https://github.com/BAMresearch/LebeDigital/tree/87-MiWoEx/lebedigital/ConcreteOntology)?
It looks to me that there are some old versions still in this folder.

eriktamsen (changes requested):
@mattheokru, @AidaZt: as already mentioned in the linked issue: please give the files readable names (if possible)
Please move the exelfile to a respective usecase or test, or if this is only used by you for development, don't include it in the PR.
I have some questions for my understanding:
1) Why do we need both drawio and rdf files for e.g. the CPTO? Should this transformation not be part of our workflow?
2) Why do we include the external otologies here? I thought, they can be accessed via a link.
3) What data format do we need for what purpose? There are drawio, owl, ttl, rdf and xml files. Sometimes the same name with different extentions, sometimes certain files in only certain formats.

soudehMasoudian (commented):
Emodule_mapping doesn't call in this branch anymore. So please approve it to merge.

joergfunger (approved):
tmp merge to be able to merge the new ontologies and then start over with the mapping in a subsequent issue

eriktamsen (approved):

joergfunger:
@soudehMasoudian  could you check for the merge conflicts and then merge

soudehMasoudian:
@joergfunger the conflict resolved and merged.

PR #95, opened 2022-12-19
94 implement youngs modulus approximation

eriktamsen:
Young's modulus function implemented with test.
Example of how to use Pint package.


lebedigital/demonstrator_scripts/youngs_modulus_approximation.py
```diff
@@ -0,0 +1,36 @@
+import numpy as np
+from lebedigital.unit_registry import ureg
+
+@ureg.wraps('MPa', ('MPa', 'kg/m^3'))
```
joergfunger, line 4: For someone not fully familiar with pint it would probably be nice to add a short documentation, both for the output (that is automatically converted to MPa and the inputs that require to use pint notation with units)

joergfunger (approved):
Very nice, just a tiny documentation request.

lebedigital/demonstrator_scripts/youngs_modulus_approximation.py
eriktamsen: Good point. I added this to the description of the function and its variables.

PR #97, opened 2022-12-19
wrapped beam design with pint package

eriktamsen:
added pint package to beam design script


lebedigital/demonstrator_scripts/beam_design.py
```diff
@@ -150,15 +153,16 @@ def beam_section_design(
             S= min(0.25 * height, 200, slmax)
     #output final design
     out = {}
-    out["top_steel_dia[mm]"] = steel_dia
+    out["top_steel_dia[mm]"] = steel_dia * ureg('mm')
```
joergfunger, line 156: isn't the steel_dia by using the decorator automatically a length unit? Same with S and As1 (based on the computations using units? So instead of adding them at the end, could we just make all the computations with units?

lebedigital/demonstrator_scripts/beam_design.py
```diff
@@ -52,6 +54,7 @@ def max_bending_moment_and_shear_force(span: float, point_load: float, distribut
             max_shear_force_dist_load + max_shear_force_point_load)
 
 
+@ureg.wraps(None, ('mm', 'mm', 'N*mm', 'N', 'N/mm^2','N/mm^2','mm','mm'))
```
joergfunger, line 57: Instead of wrapping to mm, could we just continuously use the units also in calling that function?

joergfunger: Just realized that stial_diam = int is wrong (not from your merge request)

lebedigital/demonstrator_scripts/beam_design.py
```diff
@@ -150,15 +153,16 @@ def beam_section_design(
...
-    out["shear_reinforcement_spacing[mm]"] = S
-    out["required_area_bottom_steel[mm^2]"] = As1
+    out["shear_reinforcement_dia[mm]"] = 12 * ureg('mm')
+    out["shear_reinforcement_spacing[mm]"] = S * ureg('mm')
+    out["required_area_bottom_steel[mm^2]"] = As1 * ureg('mm^2')
 
     return out
 
 
+@ureg.check('[length]','[length]','[length]','[force]','[line_load]','[stress]','[stress]','[length]',None,'[length]')
```
joergfunger, line 165: in the other merge request, you have used mass/time**2 for the line load, not sure which one is better, but I would suggest making it consistent.

lebedigital/demonstrator_scripts/beam_design.py
```diff
@@ -200,11 +208,20 @@ def check_design(span:float,
...
+    design = beam_section_design(width,
+                                 height,
+                                 max_moment,
+                                 max_shear_force,
+                                 compr_str_concrete,
+                                 yield_str_steel,
+                                 steel_dia ,
+                                 cover)
+
+    specified_area = (math.pi * (steel_dia) ** 2 / 4) * n_bottom
```
joergfunger, line 224: if area has a unit, we could remove that in the key for the dict that is returned

joergfunger (changes requested):
thanks, just a couple of remarks

lebedigital/demonstrator_scripts/beam_design.py
eriktamsen: No, using the wrapper, only checks the input and converts it if necessary.
Within the function there are no units.
You can prescribe output (c.f. youngs_modulus_approximation), but as this is a dictionary, this can not be done using the wrapper decorator.

lebedigital/demonstrator_scripts/beam_design.py
eriktamsen: Maybe, but that would require that I understand every step of the function.
I know it works with mm, so I chose to keep it that way. 
The input can still be given in cm,m,km or miles ;)

lebedigital/demonstrator_scripts/beam_design.py
eriktamsen: This is due to "work in progress" ;)
In "unit_registry.py" I defined my own line_load, as I did not find the mass/time**2 very intuitive. I will make it consistent.
Functionally it is the same.

lebedigital/demonstrator_scripts/beam_design.py
eriktamsen: Correct! Will do so :)

lebedigital/demonstrator_scripts/beam_design.py
eriktamsen: i changed it to [force]/[length], this does not require any new definition and is more intuitive

joergfunger (approved):

PR #99, opened 2022-12-20
added pint to homogenization wrapper

eriktamsen:
wrapped the fenics_concrete homogenization function with the correct units.

### DIFF
--- a/tests/simulation/test_homogenization.py
+++ b/tests/simulation/test_homogenization.py
@@ -1,5 +1,7 @@
 import pytest
 from lebedigital.simulation.concrete_homogenization import concrete_homogenization
+from lebedigital.unit_registry import ureg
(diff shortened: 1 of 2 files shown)


tests/simulation/test_homogenization.py
```diff
@@ -11,28 +13,29 @@ def test_homogenization():
     # values are kind of made up but within the expected magnitude
 
     # paste data
-    parameters['paste_E'] = 30e9  # Pa
+    parameters['paste_E'] = 30e9 * ureg('Pa') # Pa
```
joergfunger, line 16: I would personally remove the unit in the comment, but not required

joergfunger (approved):
Thanks, happy holidays.

tests/simulation/test_homogenization.py
eriktamsen: I agree. Done.

PR #101, opened 2022-12-21
100 implementing function for computation of volume contents

eriktamsen:
implemented computation of volume contents based on ratios and volume fractions, as well as the inverse to compute the fractions and ratios based on volume contents.
Both functions have tests and are using the pint package.

### DIFF
--- a/lebedigital/demonstrator_scripts/computation_volume_content.py
+++ b/lebedigital/demonstrator_scripts/computation_volume_content.py
@@ -9,6 +9,9 @@ def computation_volume_content(input_dic):
     The input and output is wrapped by the python pint package: https://pint.readthedocs.io/
     This requires units to be attached to the input values.
 
+    Volume fractions refers to the fraction of the mix volume a phase is occupying.
+    Volume content is the mass of the phase within a 1m^3.
+
     Parameters
     ----------
     input_dic : dic / pint units required converted to matching units
@@ -40,7 +43,7 @@ def computation_volume_content(input_dic):
     # initialize output dictionary
     output = {}
 
-    # optional paramter
+    # optional parameter
     if 'plasticizer_volume_content' not in input_dic.keys():
         input_dic['plasticizer_volume_content'] = 0 * ureg('kg/m^3')
         input_dic['density_plasticizer'] = 42 * ureg('kg/m^3')  # dummy value
--- a/lebedigital/demonstrator_scripts/computation_volume_content.py
+++ b/lebedigital/demonstrator_scripts/computation_volume_content.py
@@ -1,0 +1,171 @@
+from lebedigital.unit_registry import ureg
+import pytest
+
+#@ureg.wraps('MPa', ('MPa', 'kg/m^3'))
+def computation_volume_content(input) :
+    """
+    This is the function to compute volume contents based on mass or volume ratios
+
+    The input and output is wrapped by the python pint package: https://pint.readthedocs.io/
+    This requires units to be attached to the input values.
+
+    Parameters
+    ----------
+    input : dic / pint units required converted to matching units
+        required:
+            - 'density_cement' 
+            - 'density_water'
+            - 'wb_mass_ratio' : water to binder ratio water/(cement + slag (or cemII))
+            - 'density_aggregates'
+            - 'aggregates_volume_fraction'
+            - 'density_sub'
+            - 'sc_volume_fraction'
+        optional (otherwise assumed to be zero)
+             - 'plasticizer_content'
+             - 'density_plasticizer'
+        
+    Returns
+    -------
+    output : dic / pint unit given
+             - 'water_vol_fraction'
+             - 'sub_vol_fraction'
+             - 'cem_vol_fraction'
+             - 'cem_vol_content'
+             - 'sub_vol_content'
+             - 'water_vol_content'
+             - 'aggregates_vol_content'
+             - 'density_paste'
+    """
+
+    # initialize output dictionary
+    output = {}
+
+    # optional paramter
+    if 'plasticizer_content' not in input.keys():
+        input['plasticizer_content'] = 0 * ureg('kg/m^3')
+        input['density_plasticizer'] = 42 * ureg('kg/m^3')  # dummy value
+
+    # converting to correct pint units / automatic check for pint input / check for required input values
+    input['density_sub'].ito('kg/m^3')
+    input['density_cement'].ito('kg/m^3')
+    input['density_water'].ito('kg/m^3')
+    input['density_plasticizer'].ito('kg/m^3')
+    input['density_aggregates'].ito('kg/m^3')
+
+    # compute binder density
+    density_binder = input['sc_volume_fraction']*input['density_sub']+(1-input['sc_volume_fraction'])*input['density_cement']
+
+    # compute volume ratio of water to binder (cement and slag)
+    water_vol_fraction_to_binder = input['wb_mass_ratio']*density_binder/\
+                         (input['density_water']+input['wb_mass_ratio']*density_binder)
+
+    # volume fractions
+    ## plasticizer
+    output['plasticizer_vol_fraction'] = input['plasticizer_content']/input['density_plasticizer']
+    ## water
+    output['water_vol_fraction'] = (1 - input['aggregates_volume_fraction'])*water_vol_fraction_to_binder\
+                                   - output['plasticizer_vol_fraction']
+    ## binder
+    binder_vol_fraction = (1 - input['aggregates_volume_fraction'])*(1 - water_vol_fraction_to_binder)
+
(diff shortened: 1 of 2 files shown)


lebedigital/demonstrator_scripts/computation_volume_content.py
```diff
@@ -0,0 +1,185 @@
...
+    The input and output is wrapped by the python pint package: https://pint.readthedocs.io/
+    This requires units to be attached to the input values.
+
+    Parameters
+    ----------
+    input_dic : dic / pint units required converted to matching units
+        required:
+            - 'density_cem'
+            - 'density_water'
+            - 'wb_mass_ratio' : water to binder ratio water/(cement + substitute (e.g. slag or cemII))
```
joergfunger, line 18: cem II is not a substitute in itself, but if you substitute portland cement with slag or other binders, you get a cem II)

lebedigital/demonstrator_scripts/computation_volume_content.py
```diff
@@ -0,0 +1,185 @@
...
+             - 'sub_vol_content'
+             - 'water_vol_content'
+             - 'aggregates_vol_content'
+             - 'density_paste'
+    """
+
+    # initialize output dictionary
+    output = {}
+
+    # optional paramter
```
joergfunger, line 43: parmter ->parameter

lebedigital/demonstrator_scripts/computation_volume_content.py
joergfunger, line 18: are we assuming that we usually have single substitute?

lebedigital/demonstrator_scripts/computation_volume_content.py
```diff
@@ -0,0 +1,185 @@
...
+    Parameters
+    ----------
+    input_dic : dic / pint units required converted to matching units
+        required:
+            - 'density_cem'
+            - 'density_water'
+            - 'wb_mass_ratio' : water to binder ratio water/(cement + substitute (e.g. slag or cemII))
+            - 'density_aggregates'
+            - 'aggregates_volume_fraction'
+            - 'density_sub' : density of the substitute (slag or cemII)
```
joergfunger, line 21: see above

lebedigital/demonstrator_scripts/computation_volume_content.py
```diff
@@ -0,0 +1,185 @@
...
+        
+    Returns
+    -------
+    output : dic / pint unit given
+             - 'water_vol_fraction'
+             - 'sub_vol_fraction'
+             - 'cem_vol_fraction'
+             - 'cem_vol_content'
+             - 'sub_vol_content'
+             - 'water_vol_content'
```
joergfunger, line 35: What is the difference between vol fraction and vol content? Is the second one the amount with a unit for a m3 of concrete?

lebedigital/demonstrator_scripts/computation_volume_content.py
```diff
@@ -0,0 +1,185 @@
...
+    This requires units to be attached to the input values.
+
+    Parameters
+    ----------
+    input : dic / pint units required converted to matching units
+        required:
+            - 'water_vol_content'
+            - 'density_water'
+            - 'density_cem'
+            - 'cem_vol_content'
```
joergfunger, line 119: we usually would have the cem_mass in a mix documentation

tests/demonstrator_scripts/test_computation_volume_content.py
```diff
@@ -0,0 +1,138 @@
+from pint.testsuite.helpers import assert_quantity_almost_equal as assert_approx
+from lebedigital.demonstrator_scripts.computation_volume_content import computation_volume_content
+from lebedigital.demonstrator_scripts.computation_volume_content import computation_ratios
+from lebedigital.unit_registry import ureg
+
+def test_computation_volume_content() :
+    """
+    This is the test for the computation_volume_content function.
+    It test all output with  and performs a number of sanity checks
```
joergfunger, line 9: -> It tests all outputs and performs ..

tests/demonstrator_scripts/test_computation_volume_content.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+
+    # TEST No 1, check intended use in general
+    input1 = {}
+    # densities
+    input1['density_cem'] = 1.44 * ureg('g/cm^3')
+    input1['density_sub'] = 840 * ureg('kg/m^3')
+    input1['density_water'] = 977 * ureg('kg/m^3')
+    input1['density_plasticizer'] = 0.98 * ureg('g/cm^3')
+    input1['density_aggregates'] = 1.5 * ureg('t/m^3')
```
joergfunger, line 19: should we take some realistic values? The density of aggregates is usually much larger

tests/demonstrator_scripts/test_computation_volume_content.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # check all values
+    assert_approx(output1['water_vol_fraction'], 0.10510134 * ureg('dimensionless'), rtol=1e-6)
+    assert_approx(output1['sub_vol_fraction'], 0.0938778312 * ureg('dimensionless'), rtol=1e-6)
+    assert_approx(output1['cem_vol_fraction'], 0.140816747 * ureg('dimensionless'), rtol=1e-6)
+    assert_approx(output1['cem_vol_content'], 202.776115 * ureg('kg/m^3'), rtol=1e-6)
+    assert_approx(output1['sub_vol_content'], 78.8573782 * ureg('kg/m^3'), rtol=1e-6)
+    assert_approx(output1['water_vol_content'], 102.68401 * ureg('kg/m^3'), rtol=1e-6)
+    assert_approx(output1['aggregates_vol_content'], 975.0 * ureg('kg/m^3'), rtol=1e-6)
+    assert_approx(output1['density_paste'], 1126.62144 * ureg('kg/m^3'), rtol=1e-6)
```
joergfunger, line 36: Could it happen, that you input values that lead to the sanity checks performed in the library to fail? If so, such a test should be added here.

joergfunger (approved):

lebedigital/demonstrator_scripts/computation_volume_content.py
eriktamsen: I know and I know the wording is not 100% correct, but the computation is the same. 
I mean you can say you are substituting CemI for CemII.
All in all just semantics, the computation is the same.

eriktamsen: That was, what I understood. Might change when we get the data.

lebedigital/demonstrator_scripts/computation_volume_content.py
eriktamsen: see above ;)

lebedigital/demonstrator_scripts/computation_volume_content.py
eriktamsen: The volume fraction is the the fraction of the volume, like 1/3. It does not have units or m^3/m^3.
Volume content is the mass for 1m^3 of mix.
I added a comment to explain.

lebedigital/demonstrator_scripts/computation_volume_content.py
eriktamsen: this is the same

tests/demonstrator_scripts/test_computation_volume_content.py
eriktamsen: I think it is off by a factor of 2, so nothing serious. Really depends on the aggregates.
Note, the unit is given in t/m^3 (because we can :D)

tests/demonstrator_scripts/test_computation_volume_content.py
eriktamsen: I do not think so. I implemented them to be sure everything was "sane", while I was hunting some other bug. I decided I might as well leaven them there, in case for future changes.
I have not tried "stupid input", i.e. negative densities or something strange.
Assuming physically reasonable input, I think this is a robust function.

lebedigital/demonstrator_scripts/computation_volume_content.py
eriktamsen: done

lebedigital/demonstrator_scripts/computation_volume_content.py
joergfunger: So it is not a volume content (because it is a mass), but essentially the mass per concrete volume - the one that is usually specified for the mix design.

eriktamsen: well, that depends on your definition/understanding of "volume content" :D.
It is the amount of X within a defined volume. The unit is mass/length^3. 
If you think this might be confusing and have an alternative, we can easily change the name.

PR #102, opened 2023-01-06
92 erik paper branch

eriktamsen:
This is still a work in progress, but I would suggest to merge the progress of the paper into main on a regular basis, both for the chance to review the current implementation (not wording etc.), if someone is interested and mainly to decrease the chances of merge problems between Atul and me. More frequent PRs will at least reduce work needed to sort out conflicts. 

This request also includes the implementation of the paper as artifact and an update of the tensile strength function.


joergfunger (approved):
thanks

atulag0711:
Thanks Erik. I think uploading the generated pdf file also would make sense. Atleast at this stage. I will also push my WIP version. At this stage, I think we should have decoupled sections so that there is no merge conflicts. This will also foster freedom in choosing the paper writing, organizing and other git realated issues.

atulag0711 (approved):

eriktamsen:
@atulag0711 : I agree on the decoupled sections, its already setup like this.
The pdf is now included as an artifact. When you go the actions and select the respective run, you can see the pdf on the bottom.

PR #104, opened 2023-01-16
93 shacl validation

horstf:


usecases/MinimumWorkingExample/emodul/validation_test/graph.ttl
```diff
@@ -0,0 +1,37 @@
...
+    obo:RO_0010001 tt:DiamVal .
+
+tt:DiamVal a cco:InformationBearingEntity ;
+    cco:has_decimal_value 12.5 ;
+    cco:uses_measurement_unit [ a cco:MeasurementUnitOfLength ] .
+
+tt:Spec a mseo:Specimen ;
+    cco:is_input_of [ a cco:YoungsModulusTest ] ;
+    cco:designated_by [ a cco:NonNameIdentifier ] ;
+    obo:BFO_0000051 [ obo:RO_0000086 [] ] .
```
joergfunger, line 37: I guess the shape.ttl is the standard validation file for all Emodul (better Young's modulus) tests, thus it should go (and then to be extended) into LebeDigital/lebeDigital/shacl/youngs_modulus_shape.ttl. The graph.ttl that you are using is for testing purposes, thus it should be stored in LebeDigital/test/shacl/youngs_modulus.ttl

joergfunger (changes requested):
We have to make sure that the CI is running (there is currently still a problem) and there are some file movements required. The rest looks very good, thanks

lebedigital/validation.py
```diff
@@ -0,0 +1,110 @@
+from pyshacl import validate
```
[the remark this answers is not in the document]
horstf: Right, i forgot, i've renamed all files and created appropriate subfolders.

usecases/MinimumWorkingExample/emodul/validation_test/graph.ttl
horstf: I've renamed the file and created another CPTO_shape.ttl which resides in the same folder as the validation code, testing specific properties about the CPT Ontology.

tests/shacl/test_shacl.py
```diff
@@ -1,15 +1,15 @@
...
-def test_graph_against_shape():
+def test_graph_against_shacl_shape():
     """
     Testing a data graph against a shapes graph and checking that the appropriate shapes fail.
     """
-
-    shapes_location = "../usecases/MinimumWorkingExample/emodul/validation_test/shape.ttl"
-    data_location = "../usecases/MinimumWorkingExample/emodul/validation_test/graph.ttl"
+    shacl_directory = os.path.dirname(Path(__file__))
+    shapes_location = Path(shacl_directory, 'youngs_modulus_shape.ttl')
```
joergfunger, line 11: I'm a bit confused with the shapes in test and in lebeDigital. According to my understanding, there should not be a shape in test, but the shapes are defined in lebeDigital/shacl, only the example kg should be defined here. And what is not clear to me, are e planning on having a single shape file (for all potential tests, like we currently have in lebedigital/shacl/[CPTO_shacl_shape](https://github.com/BAMresearch/LebeDigital/blob/93-shacl-validation/lebedigital/shacl/CPTO_shacl_shape.ttl. or as in tests a specific one (e.g. for youngs_modulus)?

usecases/MinimumWorkingExample/dodo.py
```diff
@@ -140,69 +144,69 @@ def task_extract_processed_data_emodul():
...
-            g = validation.read_graph_from_file(graph_path)
-            res = validation.test_graph(g, shapes_graph)
-            return any(validation.violates_shape(res, shape) for shape in shapes_list)
-
-    if config['mode'] == 'cheap' or config['mode'] == 'single':
-        list_metadata_yaml_files = [ Path(metadata_emodulus_directory, single_example_name + '.yaml') ]
-    else: # go through all files
-        # list of all meta data files....
-        list_metadata_yaml_files = os.scandir(metadata_emodulus_directory)
+# #generate knowledgeGraphs
```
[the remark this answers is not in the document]
horstf: i think this is because of merge conflicts that appeared when i tried to merge main

tests/shacl/test_shacl.py
horstf: Right now the test shapes are three simple checks to see if the validation code runs as expected. Which is why there are two shapes files, one to test the code, one for the doit tasks in MinimumWorkingExample. We can change the tests to actually test the shapes themselves, instead of the code. I would need a 'correct' KG then, on which i could test the shapes.

As for the amount of shapes files, I think that maybe it would make sense to have a file per 'type' of KG (i'm lacking conceptual knowledge here though)? To prevent the shapes file from becoming too large. Right now, I've just named the shape according to the ontology file that i used to create it.

joergfunger: Ok, my suggestions would be to leave the code testing as is, and add the shapes according to the experimental type that we are defining later on. We will split the diagrams.net file into submodules (one for youngs modulus, one for mixture, ,..). So the CPTO_shacl_shape.ttl would have to be decomposed. Just to avoid to set the wrong structure, could we just exemplarily name it youngs_modulus_shacl_shape.ttl and once we have the KG generated test this approach)?

usecases/MinimumWorkingExample/dodo.py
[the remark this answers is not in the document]
horstf: i'm not sure, a git diff between origin/main and my current HEAD for dodo.py does not show these changes

tests/shacl/test_shacl.py
horstf: Yes, I've renamed the file

usecases/Concrete/dodo.py
```diff
@@ -1,10 +1,8 @@
-import graphlib
 import os
```
joergfunger, line 1: This file is going to be deleted anyway, no need to change anything here:  https://github.com/BAMresearch/LebeDigital/issues/105

PR #108, opened 2023-01-30
56 open bis integration

cezary17:
Closed the old pull request and opened this. A lot of stuff changed, namely no ExpStep anymore. Interbis stayed mostly the same + some major and minor changes to MinimumWorkingExample upload in doit.


aradermacher:
I looked into the code and did some reordering (like reformat function in doit_create_types, where it is only used) and added some stuff to the readme. @cezary17 can you please check if those things are correct.

I'm fine with the interface and looking forward to the merge :) 

Can you please rebase the two .csv files to the head before. I think we didn't changed anything there, maybe also my fault that they are there, sorry 

After the merge, I suggest the follwoing specific issues:
- create_sample_types with sections
- use json schema for predefine the metadata schema and use those for type checking before uploading new samples
- find a way to check if sample already exist in database (maybe via checksum of data files)

aradermacher (approved):

PR #109, opened 2023-01-31
Simple ontology

AidaZt:


lebedigital/SimpleOntology/Mapping/SimpleOntologyMapped.ttl
```diff
@@ -12,29 +12,29 @@ ns1: a owl:Ontology ;
 ns2:Probe_Probe1 a ns2:Artifact,
         owl:NamedIndividual ;
     ns2:has_ID ns2:ID_Probe1 ;
-    ns2:has_diameter ns2:Diameter_Probe1 ;
-    ns2:has_length ns2:Length_Probe1 ;
-    ns2:has_shape ns2:CylindricalProbe1 .
+    ns2:has_diametre ns2:Diametre_Probe1 ;
```
joergfunger, line 15: why do we change to a French version? Diameter is IMO the only correct spelling. Same with length?

joergfunger (changes requested):

joergfunger (changes requested):
the simple ontology playground is in the library, that is not correct. Please move it to a new usecase

AidaZt:
@joergfunger Do you mean to move SimpleOntology package from lebedigital to usecase package?

joergfunger (commented):
From my side OK, but the tests are not running

joergfunger (dismissed):

PR #112, opened 2023-02-03
111 add results of minimum working example to artifacts

eriktamsen:
added current results for the minimum working example to the artifacts
also removed the cheap option for the creation of the metadata, as this is quick!

### DIFF
--- a/usecases/MinimumWorkingExample/dodo.py
+++ b/usecases/MinimumWorkingExample/dodo.py
@@ -99,10 +99,10 @@ def task_extract_metadata_emodul():
     Path(metadata_emodulus_directory).mkdir(parents=True, exist_ok=True)
 
     # setting for fast test, defining the list
-    if config['mode'] == 'cheap' or config['mode'] == 'single':
-        list_raw_data_emodulus_directories = [ Path(raw_data_emodulus_directory, single_example_name) ]
-    else: # go through all files
-        list_raw_data_emodulus_directories = os.scandir(raw_data_emodulus_directory)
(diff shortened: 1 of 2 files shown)


usecases/MinimumWorkingExample/dodo.py
```diff
@@ -99,10 +99,10 @@ def task_extract_metadata_emodul():
     Path(metadata_emodulus_directory).mkdir(parents=True, exist_ok=True)
 
     # setting for fast test, defining the list
-    if config['mode'] == 'cheap' or config['mode'] == 'single':
-        list_raw_data_emodulus_directories = [ Path(raw_data_emodulus_directory, single_example_name) ]
-    else: # go through all files
-        list_raw_data_emodulus_directories = os.scandir(raw_data_emodulus_directory)
+    #if config['mode'] == 'cheap' or config['mode'] == 'single':
```
joergfunger, line 102: why is the cheap test not used anymore? If this is not required (e.g. due to the test being fast anyway), just remove the lines of code.

joergfunger (approved):

usecases/MinimumWorkingExample/dodo.py
eriktamsen: This was a suggestion of yours from some weeks ago. The extraction is fast, the slow part is the calibration/simulation.
I did not delete the lines, as I wanted the code to be there when we implement the "slower" modules.
I hope this will be done during the hackathon so this comment should not be here for long.

PR #116, opened 2023-02-09
110 building optimization workflow

eriktamsen:
I created an example workflow.


lebedigital/demonstrator_scripts/kpi_from_fem.py
```diff
@@ -0,0 +1,87 @@
...
+import pint
+import pandas as pd
+
+
+def kpi_from_fem(df,limit_temp):
+    """
+    compute KPIs from simulation output
+
+    the maximum reached temperature and the time is extracted and the difference to the limit temperature given
+    the time when the yield value equals zero is approximated
```
joergfunger, line 13: I think that needs further explanation, I guess you define a yield surface as the elastic limit based on the maximum return the maximum for any time step here, correct?

lebedigital/demonstrator_scripts/kpi_from_fem.py
```diff
@@ -0,0 +1,87 @@
...
+    df['time'] = df['time'].pint.to("seconds")
+    df['temperature'] = df['temperature'].pint.to("degree_Celsius")
+    df['yield'] = df['yield'].pint.to("dimensionless")
+
+    # subsequent operations have problems with the units (interpolation) therefore convert to standard df
+    df = df.pint.dequantify()
+
+    # adding new row with zero yield
+    new_line = pd.DataFrame({('time', 'second'): [np.nan], ('temperature', 'degree_Celsius'): [np.nan], ('yield', 'dimensionless'): [0.0]})
+    df = pd.concat([df, new_line],  ignore_index=True)
```
joergfunger, line 57: If you are sorting according to the yield, what is happening if you have two instances in time where the yield surface changes the sign? In that case you might interpolate between values of different time instances? If you remove all values after the first occurence of a neg. yield value, this would not happen, but in that case you just have to interpolate between this value and the previous one?

lebedigital/unit_registry.py
```diff
@@ -5,3 +5,7 @@
 # user defined dimensions
 ureg.define('[moment] = [force] * [length]')
 ureg.define('[stress] = [force] / [length]**2')
+ureg.define('[stress] = [force] / [length]**2')
+ureg.define('[stress] = [force] / [length]**2')
+ureg.define('GWP = [global_warming_potential] = kg_CO2_eq = kg_C02_equivalent = kg_C02eq')
```
joergfunger, line 10: see above, do we need that? And why do we have stress three times?

joergfunger (approved):
Thanks a lot, couple of remarks, but nothing serious. I think the very complex setup is very well to be followed by someone that is not familiar with the setting - that includes both the snakemake implementation, the documentation of the individual methods and the units used.

lebedigital/demonstrator_scripts/computation_GWP_mix.py
```diff
@@ -0,0 +1,43 @@
+import numpy as np
+from lebedigital.unit_registry import ureg
+
+def computation_GWP_mix(constituents):
+    """
+    This function computes the global warming potential for a square meter of cement mix
```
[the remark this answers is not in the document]
eriktamsen: you are right, fixed
technically "cement mix" might be more accurate, as we do not (yet) consider GWP of mixing/adding water etc but, we dont need to be this precise at the moment ;)

lebedigital/demonstrator_scripts/computation_GWP_mix.py
```diff
@@ -0,0 +1,43 @@
...
+    Parameters
+    ----------
+    constituents : dict
+        the dictonary expects e.g.:
+        {'cement' : {'content': 5 * ureg('kg/m^3'), 'GWP': 2 * ureg('kg_CO2_eq/kg')},
+         'aggregates': {'content': 10 * ureg('kg/m^3'), 'GWP': 1 * ureg('kg_CO2_eq/kg')}}
+
+    Returns
+    -------
+    gwp_mix : float / pint unit, will be in 'kg_CO2_eq/m^3'
```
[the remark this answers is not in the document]
eriktamsen: I see your point, but we are not actually measuring the mass of CO2. I think this makes it a bit more clear, otherwise the result will look like a density.

lebedigital/demonstrator_scripts/computation_GWP_mix.py
```diff
@@ -0,0 +1,43 @@
...
+       GWP of one cubic meter of concrete
+    """
+
+    # initialize mix gwp, with unit
+    gwp_mix = 0 * ureg('kg_CO2_eq/m^3')
+
+    # possibility to check if specific values are included
+    assert 'cement' in constituents.keys()
```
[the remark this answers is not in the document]
eriktamsen: I guess technically not.
I removed it.

lebedigital/demonstrator_scripts/computation_GWP_per_part.py
```diff
@@ -0,0 +1,38 @@
+import numpy as np
+from lebedigital.unit_registry import ureg
+
+@ureg.wraps(('kg_CO2_eq'),('kg_CO2_eq/m^3', 'm', 'm', 'm'))
+def computation_GWP_per_part(gwp_mix, width, height, length):
+    """
+    This function computes the global warming potential for a unit of beam
```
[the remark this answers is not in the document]
eriktamsen: the word unit is here not meant as "meters is a unit" but as "a piece of beam/one single beam". I changed the wording.

lebedigital/demonstrator_scripts/kpi_from_fem.py
eriktamsen: correct. I added a more detailed description

lebedigital/demonstrator_scripts/kpi_from_fem.py
eriktamsen: This case is not covered. I have not seen a simulation where that happens, as the load is constant and the material get stiffer over time.
I will add a comment to indicate that this might have to be added. but to catch this in a useful way, I would have to change this approach.

Issue #117, opened 2023-02-09
E-Module Mapping Script


alFrie:
The e-module part got separated from the rest of the CPTO. We have a working mapping script for the simple ontology and metadata to be mapped. So on this branch we will write a mapping script for the e-module. Things to do:

- [x] Edit the metadata exctraction script: Hardcode the specimen age to 28 days and add to yaml-file.
- [x] Edit the metadata exctraction script:  Extract the experiment duration ("Zeit") and add to yaml-file.
- [x] Edit the metadata exctraction script: Make the metadata-dictionary keys match the placeholders.
- [x] Map the metadata in the yaml-file to the ontology by replacing placeholders.
- [ ] Write a test.

ThiloMuth:
Test

joergfunger:
@ThiloMuth  if you accept the invitation to the repo, you can review the merge requests, e.g. [this one](https://github.com/BAMresearch/LebeDigital/pull/118)

mattheokru:
No nothing predefined by the Ontology, I will change it.

raviapatel:
> @mattheokru Is this way of spelling ("Modul") within the e-module ontology on purpose/ predefined by the ontology or something like that? It looks german to me. ![image](https://user-images.githubusercontent.com/93983136/218550519-5093f813-7e29-4037-8205-a87df04abfeb.png)
Yes this is just individual and you are right this is german way of doing it I would propose to change it to YoungsModulusTestSpecimen_ or EModulusTestSpeciemen

joergfunger:
Currently, we store the files locally (as you have mentioned with that path), so I would add for now exactly this path. Ultimately, this would have to go to a file server/mongdb/openBIS with a URI, but please talk to @AidaZt  or @ThiloMuth on how we should then reference these files in our KG.

AidaZt:
Andre suggested that we can reference the the link/URL to the file.
Example of an RDF triple:
<http://bam.de/material#experiment01> <http://bam.de/properties#rawdata> <http://bam.de/dataserver/rawdata.csv> .

joergfunger:
But that links is not existing, or how do we intend to store data such that this link is actually a real reference?

firmao:
> But that links is not existing, or how do we intend to store data such that this link is actually a real reference?

Then, we need, at least to talk about a file server, point straight to github raw files, dereferencing URIs providing RDF content, etc.

[...]

PR #118, opened 2023-02-09
117 Emodule mapping script

alFrie:
Pull request for issue #117 .


joergfunger:
Could you check the tests, they are not working.

joergfunger (changes requested):
Not sure why we do not need that anymore. Isn't it the other way round, we would have to create the knowledge graph, and then query to see if it works? What is the testfile in the current form actually doing? I guess we can already query a single entities, such as "Give me for all E-modulus tests (in that simple case we would just have one) the diameter, the reference to the files (raw data and processed data), the information whether this is a cube/cylinder and the height (in the latter case).

lebedigital/mapping/emodul_mapping.py
```diff
@@ -1,254 +1,199 @@
...
+
+
+def placeholderreplacement(
+        ontoPath,
+        metadataPath,
+        outputPath=None
+        ):
+    '''
+        Maps the values of one given metadata file (for one specimen or
+        experiment) to a given ontology, by searching within ontology linewise
```
joergfunger, line 49: not ontology, to a knowledge graph template

lebedigital/mapping/emodul_mapping.py
```diff
@@ -1,254 +1,199 @@
...
+
+
+            # ID-key is not given by metadata but created in this script, so map it now:
+            if generate_placeholder("SpecimenID") in lines[i]:
+                    logger.debug('Found placeholder "' + generate_placeholder("SpecimenID")+ '".')
+                    lines[i] = lines[i].replace(generate_placeholder("SpecimenID"), str(specimenID))
+
+
+            # depending on shape of specimen, set null pointers            
+            if "SpecimenHeight" not in keys:
```
joergfunger, line 133: This is all very specific for the Youngs modulus test. We can merge that to have something working, but as soon as we move to the next ones (mix), it would be nice to make it general. This means replace all placeholders, remove those lines with empty placeholders, and add uuid automatically (e.g. by adding a specific placeholder). Could you make an issue on that?

lebedigital/mapping/emodul_mapping.py
```diff
@@ -1,254 +1,199 @@
...
+
+    else:
+        # saving the list again to the file
+        with open(outputPath, 'w') as file:
+            for line in lines:
+                file.write(line)
+
+
+
+# # T E M P O R A R Y !!!
```
joergfunger, line 183: please remove temporary lines if merging to main

lebedigital/raw_data_processing/youngs_modulus_data/emodul_metadata_extraction.py
```diff
@@ -73,60 +70,77 @@ def extract_metadata_emodulus(rawDataPath,specimen_file,mix_file):
...
+        metadata['SpecimenDiameter'] = float(replace_comma(serviceInformation[6][1])) #diameter
+        metadata['SpecimenLength'] = float(replace_comma(serviceInformation[7][1])) #length
+        if metadata['SpecimenDiameter'] > metadata['SpecimenLength']:
             dir_name = metadata['experimentName']
             raise Exception(f'Diameter is larger then length, please fix the mistake in {dir_name}')
 
-        # set units, assuming length and diameter have the same
-        if metadata['length'] > 100:
+        # set units, assuming length and diameter have the same - This data has no placeholder yet.
+        if metadata['SpecimenLength'] > 100:
```
joergfunger, line 137: Could we just change the files and assume everything is in m (or mm)? This way, it is quite tricky to find bugs in future data.

tests/mapping/test_mapping_script.py
```diff
@@ -1,17 +1,20 @@
...
-from lebedigital.mapping.emodul_mapping import generate_knowledge_graph, get_date_time_value
+from lebedigital.mapping.emodul_mapping import placeholderreplacement
 
 
 def query_and_test_result(typeOfQuery, predicate, prop, g):
     print(f"""Query for {typeOfQuery}:""")
     q = f"""
             SELECT ?o WHERE {{
                     {predicate} {prop} ?o .
+                    FILTER (strstarts(str(), {predicate}))
```
joergfunger, line 17: What is that query supposed to do?

tests/mapping/test_mapping_script.py
```diff
@@ -27,43 +30,47 @@ def query_and_test_result(typeOfQuery, predicate, prop, g):
...
+                   'mix_file': '2014_08_05 Rezeptur_MI.xlsx'}
+
+    with open(metadataPath, 'w') as yamlFile:
+        yaml.dump(target_data, yamlFile)
+    print("Generate knowledgeGraph:")
     # generate_knowledge_graph(metadataPath, filename)
 
     # g = rdflib.Graph()
-    # result = g.parse(filename, format='ttl')
+    # g.parse(filename, format='ttl')
```
joergfunger, line 64: how to we check the generated knowledge graph? We need to have some tests that perform queries, see my general comment.

joergfunger (changes requested):
Thanks a lot, the most important change to be made is to include a reasonable query for the generated KG.

tests/mapping/test_mapping_script.py
soudehMasoudian: The thing is that we cannot use the queries Ilias wrote as the entities have ID now. For instance, we dont have "SpecimenDiameter" anymore, instead, we have "SpecimenDiameter_UniqueID". Once the format is finalized, we can change the query so that it takes the first part of the object and then fetch the result (like using "startsWith()")

tests/mapping/test_mapping_script.py
soudehMasoudian: This query should return the objects which starts with something (like diameter, height, weight, etc) as they are followed with a unique id in the current version and we cannot call them by sending the parameter directly.

joergfunger: IMHO, this type of query is not useful, since as soon as we have more than one test (with each having a diameter), this is returning list of diameters without any relation. As such, this type of query should be replaced by searching for a specific test (e.g. using the human readible id) and then return the diameter of that test.

joergfunger:
Without having looked at all the details, the tests should not use anything from the usecases, but solely rely on files in the test folder and also should not contain a sequence of files (first extract, then map, ..). The individual (intermediate) results should be hard coded.

lebedigital/raw_data_processing/youngs_modulus_data/emodul_metadata_extraction.py
joergfunger: I think Erik raised an issue with including the units in the metadata, so I guess using a single unit would be more consistent

lebedigital/mapping/emodul_mapping.py
joergfunger: I mean the mapping that you currently write just replaces in a given ttl file the corresponding strings from another dict also given as a parameter, so the input is a ttl file and a dict, the output is another ttl file (with the replacements made). IMO, this is not specific to a Youngsmodulus tests and as such could be reused for any mapping (mixture, ...).

lebedigital/raw_data_processing/youngs_modulus_data/emodul_metadata_extraction.py
eriktamsen: Jörg refers to: https://github.com/BAMresearch/LebeDigital/issues/153
I am not sure if using a single unit is the best option in the long run, but for now I am happy if it is included anywhere.
I thought it might be easier and more flexible to give every entry a unit. This includes then automatically the dimensionality of the value. Otherwise we would need to know that kind of unit is connected to which value.

lebedigital/mapping/emodul_mapping.py
joergfunger: What are the things that have to be specific? Can we just move all that to the extraction script? This would really improve the simplicity of the approach and in particular the coding requirements (because the only thing is the drawio file (and the corresponding ttl) and the json metadata structure, no python code would have to be touched.

eriktamsen:
No problem, I probably modified it at some point. I can have a look tomorrow, I am sure I can help.
What do you want it to do, compared to what it is doing now?
I am not sure if we really need this parser.
When do we need to run the script from the command line? @joergfunger, is this a required feature? Currently we are just calling the script in doit, right?

joergfunger:
The files will be automatically extracted from openBIS in the future, not sure what you mean by testfile for the mix_extraction, yaml (??). As for storing the ID, what ID are we talking about, the human readible one?

joergfunger:
Yes (I had a thumbs up on your question.

PR #123, opened 2023-02-14
Create ontoSource.owl

firmao:


joergfunger:
Every branch should have an issue first that explains what is integrated. This also means that the branch is directly related to the issue (e.g. the title with the number links to the issue). In the current setup, I'm not fully sure where to put the file, since we have a clear structure for the actual package, tests, and examples/usecases. I would expect that another directory on the main level is not what you plan on adding, so please explain via an issue what is implemented, such that this allows others to do a merge request review.

PR #132, opened 2023-02-27
fixed inputs for workflow

eriktamsen:
found the mistake and removed an outdated input to snakemake workflow


atulag0711 (approved):
Thanks Erik.

PR #143, opened 2023-03-30
split macros and bib file for paper to reduce conflicts

eriktamsen:
@atulag0711
As discussed, I tried to split all files that might lead to obvious merge conflicts.
When you start working on introduction or something similar, just let me know, or create a second introduction tex, whatever you prefer.
As discussed before, once we are moving to finalize the text, we can go to overleaf. The strength of this repo should be to create make sure all required code is publishable and useable. In addition all generated figures can be build, based on the newest raw data. this will allow us to ignore things like the designs, colors or fonts of the figures till the end, as we can always recreate them.


atulag0711 (approved):
Thanks Erik for doing this. Hopefully we don't get any nasty merge conflicts.

PR #144, opened 2023-04-03
Updating Ontologies

mattheokru:
Added placeholder for URI and updating Ontologies


lebedigital/ConcreteOntology/EModuleOntology.ttl
```diff
@@ -126,6 +126,7 @@ con:SecantModulusOfElasticty_ a con:SecantModulusOfElasticty,
     cco:has_decimal_value "$$EModule_Value$$"^^xsd:decimal ;
     cco:has_output con:ExperimentInfo_ ;
     cco:has_process_part con:Computation ;
+    cco:has_text_value "$$URI_Value$$"^^xsd:string ;
```
joergfunger, line 129: what exactly is stored here? This would then look like "URI_Value" has_text_value "URI_Value", or am I wrong?

joergfunger (changes requested):
As for the wording, I'm still in favor of naming the templates for the generation of the knowledge graph not ontology (since it is not an ontology) differently. IMO, the ontology defines the classes and properties, but usually there are no instances generated there. The KG templates that we create with diagrams.net defines instances (with placeholders), as such this is really a knowledge graph template rather than an ontology. I would recommend renaming EModuleOntology.ttl to EModuleKnowledgeGraph.ttl

lebedigital/ConcreteOntology/EModuleOntology.ttl
mattheokru: I renamed the ttl files. The Added URI value describes the testing process of the Ontology (e.g. EModule etc.). As far as I understood this value is needed to later merge the mapped ontologies? The drawio looks something like this:
Screenshot (52)

joergfunger: So please talk to @ThiloMuth, @AidaZt  @soudehMasoudian or @alFrie. IMO, we have to store the link (URI) to mix to link those, not the URI of the E-Modulus tests. The type should then also not be string, but the link to a mix. (So take you complete diagram that you have done once, remove all the links that are outside of the E-modulus box (this is what you have done to separate the tasks) and then add again this link that passes from the E-modulus to the mix. This single triple (E-modulus is_made_out_of mix, or whatever that is called).

mattheokru: Is this what you mean? The LinkedMix box contains all outside relations of the EModulus and itself is linked to a URI. I still used String as the datatype since I'm not sure what datatype a URI exactly is maybe someone knows this?
Screenshot (53)

AidaZt: for the type, can't you use something like: ^^xsd:anyURI?

mattheokru: yes that works I seem to have read over this

joergfunger: Yes, that sounds good (sorry for the late response).

mattheokru:
I added an extra Ontology for the Test Specimen, which is linked via the Specimen ID to the indsividual tests. That way not every test has to include every property of the Specimen, just the ID. Also the Mixture is now linked via the Specimen Ontology, so it is clear which Mixture Design was used to make a Test Specimen.

joergfunger:
@eriktamsen Should we modify the coverage test to exclude the ontology folder? I'm not sure how to test a new owl file.

joergfunger (approved):
I'm struggling with the "w3id,org/concrete " prefix of the placeholders, we would still have to fix the coverage test (I think we should modify the test rather than the merge request). The rest is fine with me.

eriktamsen:
As far as I can see, the coverage will be fine once the branch is updated. The ontology is not tested, it only checks python files.
@mattheokru: Please merge main into the branch, the coverage should be fine then.

eriktamsen:
@alFrie, @mattheokru: If this is ready to be merged, then the PR needs to be set to "ready for review". Currently it is still in draft mode.

PR #155, opened 2023-04-26
154 update snakemake workflow

eriktamsen:
The computation of the beam design been changed, both the output an the input
The GWP of the steel bars are added to the GWP of the beam.
The idea is to change the optimization from optimizing for aggregate content, to optimizing the beam height.
This will hopefully result in a more interesting and demonstrative example.
The snakemake workflow has been adapted for this change.

@atulag0711: I have one open question, regarding this. We no longer have a continuous variable "check_steel_area". The area is computed based on the material properties, geometry and load. There are cases where no solution is possible, i.e. too little strength of concrete, or not enough height of the beam. What would be the best way to handle this? Currently I think the script will result in `np.nan` values, which will probably make the snakemake workflow crash. I assume this might lead to problem. What would be an output of the workflow, that could work for you.


atulag0711:
@eriktamsen Thanks Erik for the updates. I have some questions/comments before I can answer your questions:

1. Now the beam design output is no of reinforcements and reinfocement diameter instead of the area check before?
2. As per your updated graph, beam design area check is no longer a constraint? So now GWP of the beam is the objective and FEM output (time and temp check) are the constraints?

Lets dissuss the nan value or the workflow crashing thing in the Wednesday meeting.

Also it would be great in the following is implemented in the workflow:
1. Some way to viz,. the beam and the relevant fields plotted over the mesh. Will be good for paper/slides.
2. A wrapper function over the snakemake workflow  which inputs the design variables and outputs objectives and the constraints. I wrote such a thing from your workflow, but you overseeing it would be better because of potential mistakes I can do.

eriktamsen:
@atulag0711: I will try to answer the questions, the rest we can do in the meeting:
1.1: Yes
1.2: basically yes, but as some combinations of fc and height will fail, is kind of is a constraint. We can discuss this in the meeting.

2.1: this exists: in branch 92:  https://github.com/BAMresearch/LebeDigital/blob/92-erik_paper_branch/usecases/optimization_paper/bam_figures/beam_design_plot.py
You can have a look at the results in the artifacts, or just run the doit workflow for the paper locally.
2.2: lets discuss this in the meeting. If you already have this, I dont see the need to reimplement this. If you include this function in a shared branch, I will be able to make changes if I need to.

atulag0711:
@eriktamsen Now the constraints are changed, so will their relationship with the Kpis(leading to constraints and objectives). Like before can you please update the analyise_kpi you had shared sometime ago? Also, I remember, temp values in the constraints were coming as NAN. Since you are at it, can you also set the dt (or any other parameter), so that we avoid getting nan in  the constraints. Nan creates a lot of problems while optimization. thanks

eriktamsen:
@atulag0711:
1) are you talking about this visualization: https://github.com/BAMresearch/LebeDigital/blob/154-update-snakemake-workflow/usecases/optimization_paper/bam_figures/beam_design_plot.py? 
Once this is merged, I will include it in my paper branch and polish the script, it currently just to make sure it works.
2) It has been a while, since I looked at that script. If I remember correctly the NaN output is for the "critical time" (when the beam can be demolded) but the required resistance is not reached. Of the top of my head I could imagine the following solutions:
1: We could increase the simulation time, this would just move the problem and significantly increase the computation time/cost
2: Instead of NaN we could set the max time of the interval. This would be the simple solution. However there would be no gradient to the values. Is this a problem?
3: I could try to do something strange, like fitting a linear function to the last data points, looking for a time outside of the simulated time frame. Not sure how the results would look like. There might be some discontinuity but I would imagine there would be a slope towards our design space.
Which options do you prefer? I will work on that in a separate issue, so if you are fine with the beam design changes, we can merge this PR.

atulag0711:
> @atulag0711:
> 
> 1. are you talking about this visualization: https://github.com/BAMresearch/LebeDigital/blob/154-update-snakemake-workflow/usecases/optimization_paper/bam_figures/beam_design_plot.py?
>    Once this is merged, I will include it in my paper branch and polish the script, it is currently just there to make sure it works.
> 2. It has been a while, since I looked at that script. If I remember correctly the NaN output is for the "critical time" (when the beam can be demolded) but the required resistance is not reached. Of the top of my head I could imagine the following solutions:
>    1: We could increase the simulation time, this would just move the problem and significantly increase the computation time/cost
>    2: Instead of NaN we could set the max time of the interval. This would be the simple solution. However there would be no gradient to the values. Is this a problem?
>    3: I could try to do something strange, like fitting a linear function to the last data points, looking for a time outside of the simulated time frame. Not sure how the results would look like. There might be some discontinuity but I would imagine there would be a slope towards our design space.
>    Which options do you prefer? I will work on that in a separate issue, so if you are fine with the beam design changes, we can merge this PR.

1. I was talking about https://github.com/BAMresearch/LebeDigital/tree/89_Calibration_Optimisation_HydrationSolver/usecases/optimization_paper/analyze_kpis. For optimization, it is very important for me to understand, how the kpis depend on the design variable. Essentially the workflow is for this reason only to my understaning, that it seamlesssly connect design variables to the KPIs. This kind of function (an option to viz is a plus) would be great. 
 
 
 I think the second solution should work.  If not we can discuss it later. thanks Erik.

eriktamsen:
I understand now, however that has nothing to do with the current PR. I will think of something similar for the visualization, when I am polishing that part of the paper.

eriktamsen:
I created a new Issue for imporving the FEM KPIs: https://github.com/BAMresearch/LebeDigital/issues/165

atulag0711:
> I understand now, however that has nothing to do with the current PR. I will think of something similar for the visualization, when I am polishing that part of the paper.

Ahhh. maybe I did a bad job in communicating, I didnt talk about the paper. I seek the data/plot to understand better the admisisble region of the design variables. I need to understand how the updated design variables influence the objectives and constraints to perform any meaningful optimization.  If we are not on the same page, we can do a short call anytime after 3.30pm today? let me know.

atulag0711 (approved):
Thanks for doing this. please dont forget to later add a figure/table which plots inputs (design variable cem ratio and beam height) and the output (objectives and constraints)

Something like this: https://github.com/BAMresearch/LebeDigital/blob/89_Calibration_Optimisation_HydrationSolver/usecases/optimization_paper/analyze_kpis/kpis.csv

PR #162, opened 2023-05-08
Uploaded the CPTO Ontolgy

mattheokru:


joergfunger (changes requested):
The outdated version are redundant, could you just add the current version and remove all others? Otherwise, we never know what is really correct. The history is also stored in git, this if you want to retrieve an old version, this can be easily done be checking out an older version of the repo.

joergfunger (dismissed):
What is the difference between CPTO_Ontology.ttl and Concrete_Ontology_MSEO.owl. And 
MergedAllCoreOntology-v1.3.ttl is that the old pmd core?

mattheokru:
The Concrete_Ontology_MSEO is the Ontology that contains the added Classes and Properties. This is our "Concrete Ontology". It includes the MSEO but still needs the MergedAllCoreOntology to work (ConcreteOntology needs to import MergedOntology). The CPTO_Ontology is the Ontology that contains all the Individuals and relations for the MiWoEx and is using the Concrete_Ontology_MSEO

joergfunger:
What do you mean by "The CPTO_Ontology is the Ontology that contains all the Individuals and relations for the MiWoEx"? The individuals are created in the KG-templates usign diagrams.net and chowlk. Or what individuals are you referring to?

mattheokru:
Yes the CPTO_Ontology is just the exported KG of the MiWoEx without data

joergfunger:
But that is not used, correct? We use only the ontology, and then the individual knowledge graphs that are then put together?

mattheokru:
Yes excatly, so it isn't really needed. But I thought it would be easier to write a mapping script if there is a basic example of what it will look like, once the KG gets exported. If that isn't necessary I will delete it

joergfunger:
Could you give me a call the next time you work on that such that we could fix that faster.

PR #166, opened 2023-05-12
122 openbis jsonschema for typechecking

cezary17:
Added a type-checker for openbis samples based on pydantic models/validators. Type-checkers can also in some cases automatically reformat the data to make it openbis-conforming (mostly datetime formats). Any suggestions welcome.


lebedigital/openbis/interbis.py
```diff
@@ -659,3 +682,70 @@ def combine_urls(base_url, relative_url):
...
+    def _get_datatype_conversion(self, property_name: str, property_datatype: str):
+        """
+        Converts the openbis datatypes into python datatypes if possible, else a custom datatype
+        Use with the `generate_typechecker` method
+        """
+        # if the prop is in the dict then it is not a CONTROLLED_VOCABULARY
+        if property_datatype in CONVERSION_DICT:
+            return CONVERSION_DICT[property_datatype]
+
+        vocabulary_name = self.get_property_type(property_name).vocabulary
```
joergfunger, line 695: what exactly is happening here, could you document that answer as well?

tests/openbis_functions/test_interbis.py
```diff
@@ -145,9 +191,9 @@ def setup(pytestconfig):
     o.logout()
 
 
-@pytest.mark.login
-@pytest.mark.parametrize("write, sheet_name, path", [(False, "metadata", ""),
-                                                     (True, "some_named_sheet", Filepaths.excel_output.value)])
+@ pytest.mark.login
```
joergfunger, line 194: why did you remove the space?

tests/openbis_functions/test_interbis.py
```diff
@@ -164,7 +210,7 @@ def test_get_metadata_import_template(setup, pytestconfig, expected_df_import, w
         pd.testing.assert_frame_equal(df, expected_df_import)
 
 
-@pytest.mark.login
+@ pytest.mark.login
```
joergfunger, line 213: same here, without space?

joergfunger (approved):
Minor remarks, can be merged after those are adressed

tests/openbis_functions/test_interbis.py
aradermacher: at least do it consistently for the whole file

usecases/MinimumWorkingExample/upload_scripts/doit_upload.py
```diff
@@ -276,16 +276,12 @@ def _read_metadata(yaml_path: str, sample_type_code: str, default_props: dict):
...
     for key, val in loaded.items():
-        if val is None:
-            continue
         if key in default_keys:
-            if key.lower() == 'operator_date':
-                # convert german date to openBIS date format YYYY-MM-DD
-                data[key] = parse(val).strftime('%Y-%m-%d')
-            else:
-                data[key] = val
+            data[key] = val
```
aradermacher, line 284: nice!

usecases/MinimumWorkingExample/upload_scripts/doit_upload.py
```diff
@@ -478,6 +470,10 @@ def _mixture_upload(
         parents=list(mixture_parent_ingredients.keys()),
     )
 
+    MixtureModel = o.generate_typechecker(mixture_sample_type)
+    mixture_model_return = MixtureModel(**mixture_metadata_dict)
+    mixture_metadata_dict = mixture_model_return.dict(exclude_unset=True)
```
aradermacher, line 475: Please add the use of the typechecker in teh readme of "interbis modul"

usecases/MinimumWorkingExample/upload_scripts/doit_upload.py
```diff
@@ -588,6 +585,11 @@ def _emodul_upload(
     logger.debug(pformat(emodul_metadata_dict))
     # logger.debug(o.get_sample_type_properties(emodul_sample_type))
     # Setting the metadata from the yaml file metadata, setting '$name' to 'experimentName'
+
+    EmodulModel = o.generate_typechecker(emodul_sample_type)
+    emodul_model_return = EmodulModel(**emodul_metadata_dict)
+    emodul_metadata_dict = emodul_model_return.dict(exclude_unset=True)
```
aradermacher, line 591: Would it be possible to use the typechecker also for object type dataset?

aradermacher (approved):
Nice work! Thanks!
Please update the readme before merging

lebedigital/openbis/interbis.py
```diff
@@ -30,9 +48,11 @@ def __dir__(self):
             "get_collection_identifier()",
             "exists_in_datastore()",
             "create_sample_type()",
-            "create_parent_hint()"
-            "set_parent_annotation()"
-            "get_parent_annotation()"
+            "create_parent_hint()",
```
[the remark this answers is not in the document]
cezary17: As far as I know the `dir()` method is there to show/return the attributes and methods of an object. It is defined in pybis in the openbis class and seeing as this code is extending that class I have also extended that list. My guess as to why they (pybis developers) have implemented it is to hide some internal functions from the user as python does not do public/private functions.

lebedigital/openbis/interbis.py
cezary17: I added a comment in that spot. What is happening is that we can generally just simply cast props to their python types. For example the types VARCHAR, REAL or INTEGER to string, int and float respectively so a simple dict lookup to a static dict is enough for pydantic to know what it needs to get. Controlled Vocabularies are the only data type that is not in that dict because it is the only data type which needs special treatment - constructing an Enum with their accepted values which we then run against our generated/read data we want to upload to openbis.

tests/openbis_functions/test_interbis.py
cezary17: I have no idea what happend in those decorators, must have been a formatter error. I havent added random spaces by hand. I have removed the spaces.

tests/openbis_functions/test_interbis.py
cezary17: Same as above

usecases/MinimumWorkingExample/upload_scripts/doit_upload.py
cezary17: I have extended the readme. Can you check it if it is enough for the potential users?

cezary17:
I have implemented the requested changes and resolved the conversations. If you can take a look if the changes are satisfactory then I would appreciate it. @joergfunger @aradermacher

lebedigital/openbis/README.md
```diff
@@ -162,4 +162,22 @@ samples = o.get_samples(
...
+SampleModel = o.generate_typechecker("SAMPLE_TYPE")
+sample_model_return = SampleModel(**sample_properties)
+sample_properties = sample_model_return.dict(exclude_unset=True) # exclude_unset is for not including other property types that were not set in sample_properties
+```
+
+> The sample_properties have been checked for their types and casted/formatted if possible. The functionality includes:
+
+- Casting strings like "2.13" or "7" to float or integer respectively
+- Checking if CONTROLLED_VOCABULARY Properties are withing their defined vocabularies
+- Casting date-strings into the required format. Ex "17.10.2023 10:45" will be formatted to "2023-10-17"
```
aradermacher, line 183: Thanks for adding that.
What's with the time? Shouldn't it be 2023-10-17 10:45 at the end?

aradermacher (approved):

lebedigital/openbis/README.md
cezary17: Not if the property type is a date, if it was a timestamp then yes it would include the date. The formatting itself is in my opinion not that important though as the main point is that the formatting will be done automatically.

PR #172, opened 2023-05-22
165 improve fem kpi

eriktamsen:
I hope this is what you where looking for Atul.


atulag0711 (approved):
Thanks Erik. Please merge.  I will merge main and run the optimization problem for this. Just for the record here,
1. As I understand the "workflow" is not crashing in the range of design variables.
2. The FEM however crashes in some part of the range, but I am told it will not affect the kpis somehow.

PR #180, opened 2023-06-14
add json package

eriktamsen:
fixed problem with tests


joergfunger (approved):
I all reviews were that short :)

PR #183, opened 2023-06-21
fix tests

eriktamsen:
had to change the expected values for some tests as the yield values from fenics_concrete have been normalized


joergfunger (approved):
I guess you changed completely the setup, because the result seem to be quite different, correct?

eriktamsen:
@joergfunger, yes and no. Its still the same computation, only the "yield values" are divided by the tensile or compressive strength depending which constraint is triggered.

joergfunger:
yes, I meant just the results are completely different (even though scaling is a small step for the files, but one giant leap for the test files) taken from Armstrong when landing on the moon :)

PR #185, opened 2023-06-22
fixing seed for reproducibility

atulag0711:


lebedigital/calibration/calibrationWorkflow.py
```diff
@@ -117,6 +117,10 @@ def estimate_youngs_modulus(
     low_sigma = 0
     high_sigma = 0.005
 
+    # for reproducible results
+    seed =1
```
eriktamsen, line 121: This will use the same seed every time also for the real calibration.
This can't be a good solution.
What @joergfunger was asking, is to fix the seed only for the tests. This would require a possibility to pass the seed in the interface.

eriktamsen (changes requested):
@atulag0711: Thank you for dealing with this quickly. However, for what I understand, this is not a useful implementation.
We want the seed to be fixed in the test but otherwise have a "real" random behavior.

lebedigital/calibration/calibrationWorkflow.py
atulag0711: Alright updated. please pull whenever the tests pass.

eriktamsen (approved):
Thank you for the changes @atulag0711

PR #188, opened 2023-07-05
187 fixing breaking tests

cezary17:
Should work now. pydantic has some breaking changes with versions >2.0 and something(?) broke in pybis 1.35.6 which caused the issues. I've locked both of them to the last working versions for now.


eriktamsen (approved):
Looks good for now. Thanks for the quick fixes.

PR #190, opened 2023-07-10
Knowledge graphs update

mattheokru:
Merging the updated Knowledge Graphs into main


lebedigital/ConcreteOntology/BaseOntologies/Taxonomie.xlsx
eriktamsen, line 1: Would it be possible to use a different format? Maybe csv?
Otherwise its difficult to use/view.

lebedigital/ConcreteOntology/BaseOntologies/tax2ont/Taxo.xlsx
eriktamsen, line 1: Would it be possible to use a different format? Maybe csv?
Otherwise its difficult to use/view.

lebedigital/ConcreteOntology/BaseOntologies/tax2ont/main.py
eriktamsen, line 1: I am not sure how/when this script is used/called.
In general we try to define flexible function within the python scripts and then these can be called from a workflow/usecase. Is that possible here?

lebedigital/ConcreteOntology/BaseOntologies/tax2ont/main.py
```diff
@@ -0,0 +1,126 @@
...
+PrefixDef = "http://www.w3.org/ns/prov#definition"
+# Prefix of the Annotation Property used for definition source
+PrefixDefSource = "https://w3id.org/pmd/co/definitionSource"
+NameExcelFile = "Taxo.xlsx"
+
+
+# welcome message
+print("Hi, this tool is for converting a taxonomy from an Excel file to an ontology.")
+# Eingabe Aufforderung
+wahl = input('Do you want to add to a already existing Ontology (Press "1") or use a blank template (Press "2")?')
```
eriktamsen, line 20: so this is used "manually" in the process of building an ontology, not a function that needs to be integrated into an automated workflow?

usecases/demonstrator/ExperimentData/Examples/extractor.py
eriktamsen, line 1: I am not 100% sure what is going on.
What is the purpose of this script?
The name says its an example. But it seems to be hard coded for some specific purpose, not the example of how to use a flexible function.
If I understand correctly, I would suggest to write a flexible function, add that to the "lebedigital" directory and then write an "exaple" test for the test directory, but maybe the purpose is something different.

usecases/demonstrator/ExperimentData/Examples/processed_data/Sample - Ch 1_calo.csv
eriktamsen, line 1: The processed data should not be committed.
There should be a workflow (e.g. a dodo file) that automatically generates the processed files.

eriktamsen (dismissed):
Thanks for the pull request.
Its good to get some updates into the main branch from time to time.
The generation of the KG is then the next thing that will be integrated?

1) Just a small comment for the future, please create an issue first, then make a branch, that way all branches are numbered and linked to an issue, making the overview easier.

2) Why do the "BaseOntologies(old)" and "CPTO(old)" directories still exist? With git, you can go "back in time" and look at older versions.

3) Please make sure processed and generated data is not committed, but automatically generated by a respective workflow. How are the "processed_data" files created?

usecases/demonstrator/ExperimentData/Examples/extractor.py
AidaZt: this script is from Ravi, we edited the script.  it contains now a function that can be called for the extraction, similar to the other extraction scripts. [https://github.com/BAMresearch/LebeDigital/blob/117-e-module-mapping-script/lebedigital/raw_data_processing/Heat_of_Hydration/HoH_metadata_extraction.py](url) , we would use tis script instead of extractor.py

eriktamsen: Sounds good. As the link does not work, I am not sure what you are planning. 
1) Just make sure al used functions are located in "lebedigtial" not "usecases"
2) there are tests for these functions in "tests"
3) that old files like "extractor.py" are not commited in a PR to main

lebedigital/ConcreteOntology/BaseOntologies/tax2ont/main.py
mattheokru: @eriktamsen sorry for the mess. I forgot to delete the files I used for creating the Ontology and the old BaseOntologies. I removed all unnecessary files. Also the Taxonomie is now in .csv

usecases/demonstrator/ExperimentData/Examples/extractor.py
eriktamsen, line 1: I think this script needs to be moved to the lebedigital folder, we try not to have scripts in the usecases.
It would be good to make it more general, write a test and start building a workflow for the demonstrator, that includes the data extraction like this.

usecases/demonstrator/ExperimentData/Examples/extractor.py
```diff
@@ -0,0 +1,48 @@
+import json as js
+import os
+import numpy as np
+with open("calo_example.csv","r") as f:
```
eriktamsen, line 4: As mentioned above, this script is a hard coded example for a single file.
This does not seem practical.
Rewrite it as a funciton that take the flie as input, might be already sufficient.

usecases/demonstrator/ExperimentData/Examples/processed_data/Sample - Ch 1_calo.csv
eriktamsen, line 1: This is processed data (according to the name of the directory).
This should not be commited, but generated by the workflow.

eriktamsen (commented):
Thank you for the changes.
The last point to imporve is related to the extractor.py script, see my comments.

usecases/demonstrator/ExperimentData/Examples/processed_data/Sample - Ch 1_calo.csv
eriktamsen: it looks to me as if this is all regarding the heat of hydration.

PR #193, opened 2023-07-17
192 check workflow constraints

eriktamsen:
improved constraint

### DIFF
--- a/lebedigital/demonstrator_scripts/dummy_paste_strength_stiffness.py
+++ b/lebedigital/demonstrator_scripts/dummy_paste_strength_stiffness.py
@@ -30,8 +30,8 @@ def dummy_paste_strength_stiffness(slag_ratio, phi_paste):
         paste_youngs_modulus_min + (paste_youngs_modulus_max - paste_youngs_modulus_min) * slag_ratio
     )
 
-    paste_strength_min = 25 * ureg("MPa")
+    paste_strength_min = 5 * ureg("MPa")
     paste_strength_max = 40 * ureg("MPa")
-    paste_strength = paste_strength_min + (paste_strength_max - paste_strength_min) * slag_ratio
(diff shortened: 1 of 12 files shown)


lebedigital/demonstrator_scripts/dummy_paste_strength_stiffness.py
```diff
@@ -30,8 +30,8 @@ def dummy_paste_strength_stiffness(slag_ratio, phi_paste):
         paste_youngs_modulus_min + (paste_youngs_modulus_max - paste_youngs_modulus_min) * slag_ratio
     )
```
atulag0711, line 32: E also needs to be updated right? similar to the paste_strength i.e., 
paste_youngs_modulus = (
        paste_youngs_modulus_max - (paste_youngs_modulus_max - paste_youngs_modulus_min) * slag_ratio

tests/demonstrator_scripts/test_dummy_scripts.py
```diff
@@ -16,4 +18,4 @@ def test_dummy_scripts():
 
     E, fc = dummy_paste_strength_stiffness(0, 10)
     assert E.magnitude == pytest.approx(30)
```
atulag0711, line 20: I guess then accordingly this will change.

atulag0711 (approved):
Thanks Erik. Just minor changes in the dummy E script for convenience.

lebedigital/demonstrator_scripts/dummy_paste_strength_stiffness.py
eriktamsen: Needs is a strong word :D, this is a dummy script but by implementing the change the trend might be closer to what we would expect in real life.
I made the change as requested.

tests/demonstrator_scripts/test_dummy_scripts.py
eriktamsen: Changed accordingly

PR #200, opened 2023-09-01
reset pandas to older version, probably conflict with  pint-pandas

joergfunger:
Pandas is reset to 2.0.3, there are apparently some conflict with pint-pandas


srosenbu (approved):
Looks good ;)

PR #208, opened 2023-11-17
207 remove interbis from lebedigital

cezary17:
Removed Interbis from Lebedigital and made a separate Package called ExtPybis that was uploaded to the bam77 conda channel. 

I don't think I can do anything to increase coverage because I deleted now redundant tests for Interbis and that decreased the overall coverage.


cezary17:
The coverage of Interbis was 90%, overall coverage of current main is 85%, after deleting Interbis (and the Interbis tests) it is 84%

joergfunger (approved):
Thanks, looks good to me. I'll try to force push to main due to the coverage decrease.

joergfunger:
Force pushed due to coverage increase (due to the src code that was transferred having a larger coverage, the overall coverage decreased)

PR #220, opened 2024-02-19
Remove coverage badge

srosenbu:
I removed the coverage badge and the related github workflow which failed because it required data from Eriks personal github account.


joergfunger (changes requested):
I personally feel that the code should directly be removed instead of commented out. If someone wants to use that they could use the log message to see when the badge has been removed.

